PR #24, opened 2022-11-09
VulkanVideoParser: Fix maxActiveReferencePictures validation error

dabrain34:
According to
https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkVideoSessionCreateInfoKHR-maxActiveReferencePictures-04849

the maxActiveReferencePictures should not exceed 16

### DIFF
--- a/vk_video_decoder/libs/VkVideoParser/VulkanVideoParser.cpp
+++ b/vk_video_decoder/libs/VkVideoParser/VulkanVideoParser.cpp
@@ -44,7 +44,7 @@ static const uint32_t bottomFieldMask = (1 << bottomFieldShift);
 static const uint32_t fieldIsReferenceMask = (topFieldMask | bottomFieldMask);
 
 static const uint32_t MAX_DPB_REF_SLOTS = 16;
-static const uint32_t MAX_DPB_REF_AND_SETUP_SLOTS = MAX_DPB_REF_SLOTS + 1; // plus 1 for the current picture (h.264 only)
+static const uint32_t MAX_DPB_REF_AND_SETUP_SLOTS = MAX_DPB_REF_SLOTS; // plus 1 for the current picture (h.264 only)
 
 #define COPYFIELD(pout, pin, name) pout->name = pin->name
 


dabrain34:
Is there a good reason to have this line https://github.com/nvpro-samples/vk_video_samples/blob/8c79dc297c33d92f6fe9cb573beb99fc9657c74e/vk_video_decoder/libs/VkVideoParser/VulkanVideoParser.cpp#L47
causing the Validation Layer error.

dabrain34:
@zlatinski

zlatinski:
> the maxActiveReferencePictures should not exceed 16

Well, [dabrain34](https://github.com/dabrain34), the issue is that h.264 supports up-to 16 references and we need one extra for the setup slot (the reference to be). So, essentially we may end-up having up-to 16 + 1 active references.

From the spec:
> maxActiveReferencePictures is the maximum number of [active reference pictures](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#active-reference-pictures) a single video coding operation can use.

We still have maximum of 16 active reference pictures, because only 16 of the 17 dpb slots are active (are used as reference) at the same time. After the decode operation, one can still only use 16 DPBs as references.

Where exactly is the validation layer failing?

charlie-ht:
I hit this validation error in the CTS as well and posted a CL to fix it there. I'm not able to use the video samples here on my hardware with the Intel drivers due to crashes, so I can't test the state here, but some comments since the CTS decoder was mostly copied from an older version of this repo.

The confusion there stemmed from using maxDpbSlotCount for both maxDpbSlots and maxActiveReferencePictures. The value should be checked against the respective video capabilities of the selected video profile, but they were not in the case of the CTS. While it was legal and necessary to have 16+1 for the maxDpbSlots, it is not for the maxActiveReferencePictures, which was the root of the given error.

zlatinski:
> While it was legal and necessary to have 16+1 for the maxDpbSlots, it is not for the maxActiveReferencePictures, which was the root of the given error.

Thank you, Charlie! maxActiveReferencePictures max value is fixed now at the ToT.

charlie-ht:
@zlatinski We can close this one now :)

zlatinski:
Closing this issue as fixed.

PR #31, opened 2023-02-12
Enabled AMD Path.

Sonuthomas-AMD:
Tested on AMD NV1x and NV2x
Tested on NVIDIA Ampere.
Tested H264 and HEVC(8 & 10bit)
Signed-off-by: Thomas, Sonu <sonu.thomas@amd.com>


zlatinski:
Thank you, for the change, Sonuthomas! I'm looing at the change and will merge it soon.

vk_video_decoder/libs/VkCodecUtils/VulkanDeviceContext.cpp
```diff
@@ -167,17 +167,18 @@ bool VulkanDeviceContext::HasAllDeviceExtensions(VkPhysicalDevice physDevice, bo
...
-            if (printMissingExt) {
-                std::cerr << "HasAllDeviceExtensions() ERROR: requested device extension "
-                    << name << " is missing!" << std::endl << std::flush;
-            }
-            return false;
+    auto nameIterator = m_reqDeviceExtensions.begin();
+    while (nameIterator != m_reqDeviceExtensions.end()) {
+        if (ext_names.find(*nameIterator) == ext_names.end()) {
+            std::cout << " Extension " << *nameIterator << " is not supported" << std::endl;
+            nameIterator = m_reqDeviceExtensions.erase(nameIterator);
```
zlatinski: We can't modify this list - this is the user requested list of extensions. If we can't satisfy the extensions list, it needs to be adjusted. What extension(s) was/were causing problem for AMD?

vk_video_decoder/libs/NvCodecUtils/FFmpegDemuxer.cpp
```diff
@@ -379,14 +379,14 @@ class FFmpegDemuxer : public VideoStreamDemuxer {
...
-        // assert(!"Unknown CHROMA_SUBSAMPLING!");
+        assert(!"Unknown CHROMA_SUBSAMPLING!");
         return VK_VIDEO_CHROMA_SUBSAMPLING_444_BIT_KHR;
     }
 
     virtual uint32_t GetProfileIdc() const
     {
-        // return (uint32_t)profile;
-        return STD_VIDEO_H264_PROFILE_IDC_MAIN;
+         return (uint32_t)profile;
```
zlatinski, line 388: Nice touch.

vk_video_decoder/include/vkvideo_parser/VulkanVideoParser.h
```diff
@@ -35,7 +35,7 @@ class IVulkanVideoDecoderHandler : public VkVideoRefCountBase {
                                       const uint8_t* pInitializeBufferMemory,
                                       size_t initializeBufferMemorySize,
                                       VkSharedBaseObj<VulkanBitstreamBuffer>& bitstreamBuffer) = 0;
-
+    virtual bool IsDstDpbDistinctImages() = 0;
```
zlatinski, line 38: I don't think we need this additional function - please see my comments about DPB/output images.

vk_video_decoder/libs/VulkanVideoFrameBuffer/VulkanVideoFrameBuffer.cpp
```diff
@@ -1063,6 +1112,20 @@ int32_t NvPerFrameDecodeImageSet::init(const VulkanDeviceContext* vkDevCtx,
             if (result != VK_SUCCESS) {
                 return -1;
             }
+            if (pDpbImageCreateInfo != nullptr) {
```
zlatinski: I'm not sure why do we need all this code for re-indexing.
I also don't think we need the additional m_dpbImageView and all the complexity associated with this additional image.

Here is the idea about how the application is structured:

There are two images in NvPerFrameDecodeResources. m_frameImageView and m_displayImageView.

m_frameImageView: is a both OUT and DPB image for the COINCIDE (NV and intel) case. It is only DPB for the DISTINCT (AMD) case.

m_displayImageView is the decoder's OUT image for DISTINCT. It is also the NV out image when the output is required to be linear (non-optimal).  I would suspect AMD supports linear output for the OUT decoded images, correct?

So, I think, following this logic,  we should not need a third image.

At the decoder's side, when we are getting the out and DPB images, the configuration of the NvPerFrameDecodeResources should know what to return in NvPerFrameDecodeResources::GetImageSetNewLayout().

in the AMD's case it should return the separate images in pDpbPictureResource and pOutputPictureResource. For NV and Intel it should return the same image for both.

Does this make sense? I don't mind renaming m_frameImageView to  m_outDpbImageView and m_displayImageView to m_outImageView, BTW.

zlatinski (changes requested):
Hi Sonu,
Thank you very much, again, for the change!

The change, in general, looks good to me, but there a few minor thing that need to be addressed and also DPB/OUT image restructuring (please see below).

I would suggest you break this big change to smaller changes. For example your changes to profileIdc and required extensions can go as separate patches. Let's have the DPB/OUT image changes as a separate change, as well.

Thank you!
Tony

vk_video_decoder/libs/VkCodecUtils/VulkanDeviceContext.cpp
Sonuthomas-AMD: VK_EXT_YCBCR_2PLANE_444_FORMATS_EXTENSION_NAME
VK_KHR_VIDEO_ENCODE_QUEUE_EXTENSION_NAME

These are not supported by AMD.

zlatinski: > VK_EXT_YCBCR_2PLANE_444_FORMATS_EXTENSION_NAME VK_KHR_VIDEO_ENCODE_QUEUE_EXTENSION_NAME
> 
> These are not supported by AMD.

We can remove these extensions - NP.

vk_video_decoder/libs/VkVideoDecoder/VkVideoDecoder.cpp
```diff
@@ -693,7 +686,7 @@ int VkVideoDecoder::DecodePictureWithParameters(VkParserPerFrameDecodeParameters
     VkVideoEndCodingInfoKHR decodeEndInfo = { VK_STRUCTURE_TYPE_VIDEO_END_CODING_INFO_KHR };
     m_vkDevCtx->CmdEndVideoCodingKHR(frameDataSlot.commandBuffer, &decodeEndInfo);
 
-    if (m_useSeparateOutputImages || m_useLinearOutput) {
+    if ((!m_useSeparateOutputImages) && m_useLinearOutput) {
         CopyOptimalToLinearImage(frameDataSlot.commandBuffer,
```
zlatinski, line 690: if (m_useSeparateOutputImages || m_useLinearOutput)
This above condition should stay the way it is, but add a third one for the the OUTPUT_DISTINCT;
if (!outputDistinct && (m_useSeparateOutputImages || m_useLinearOutput)) {
...
}

vk_video_decoder/libs/VkVideoDecoder/VkVideoDecoder.cpp
```diff
@@ -208,7 +208,14 @@ int32_t VkVideoDecoder::StartVideoSequence(VkParserDetectedVideoFormat* pVideoFo
     }
 
     m_capabilityFlags = videoDecodeCapabilities.flags;
+    if (IsDstDpbDistinctImages()) {
+        m_useSeparateOutputImages = true;
+    }
```
zlatinski, line 214: We still nee the useLinearOutput case here:

if (IsDstDpbDistinctImages() || useLinearOutput ) {

vk_video_decoder/libs/VkVideoDecoder/VkVideoDecoder.cpp
```diff
@@ -549,6 +537,11 @@ int VkVideoDecoder::DecodePictureWithParameters(VkParserPerFrameDecodeParameters
             assert(imageBarriers[numDpbBarriers].image);
             numDpbBarriers++;
         }
+        }
+    if ((m_capabilityFlags & VK_VIDEO_DECODE_CAPABILITY_DPB_AND_OUTPUT_DISTINCT_BIT_KHR)) {
+        pPicParams->pictureResources[pPicParams->numGopReferenceSlots] = pPicParams->decodeFrameInfo.dstPictureResource;
+        pPicParams->decodeFrameInfo.dstPictureResource = currentOutputPictureResource;
+
     }
```
zlatinski, line 545: I don't think this function would need any modifications anymore, right?

zlatinski (commented):
The changes look good, in general. I still have  a few minor comments and there are few places with weird formatting. Let me start pushing these changes with a bit of formatting one after another until we are all happy with the outcome ...

vk_video_decoder/libs/VkVideoDecoder/VkVideoDecoder.cpp
Sonuthomas-AMD: i think this change is required

zlatinski:
Hi Sonu,
Please check out [distinct-dpb-and-output-wip](https://github.com/nvpro-samples/vk_video_samples/tree/distinct-dpb-and-output-wip).

PR #37, opened 2023-04-28
vk_video_decoder: fix segv with invalid args

dabrain34:
Fix crash with ./vk-video-dec-test -h (to get the help) Fix other use case where atoi is used with a null string.


zlatinski (approved):
Thank you, for the fixes, Stéphane!

zlatinski:
Merged - thank you!

Issue #38, opened 2023-04-28
vk_video_encoder: mismatch usage of vulkan.hpp during app build


dabrain34:
As a first error, I got: 
```
In file included from /DEV/nvpro_samples/vk_video_samples/vk_video_encoder/../common/include/vulkan/vulkan.hpp:5660,
                 from /DEV/nvpro_samples/vk_video_samples/vk_video_encoder/VkCodecUtils/NvPictureBuffer.h:21,
                 from /DEV/nvpro_samples/vk_video_samples/vk_video_encoder/VkCodecUtils/NvPictureBuffer.cpp:17:
/usr/include/vulkan/vulkan_to_string.hpp: In function ‘std::string vk::to_string(vk::QueryPipelineStatisticFlags)’:
/usr/include/vulkan/vulkan_to_string.hpp:512:50: error: ‘eClusterCullingShaderInvocationsHUAWEI’ is not a member of ‘vk::QueryPipelineStatisticFlagBits’
  512 |     if ( value & QueryPipelineStatisticFlagBits::eClusterCullingShaderInvocationsHUAWEI )
      |                                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/vulkan/vulkan_to_string.hpp: In function ‘std::string vk::to_string(vk::ShaderStageFlags)’:
/usr/include/vulkan/vulkan_to_string.hpp:890:39: error: ‘eClusterCullingHUAWEI’ is not a member of ‘vk::ShaderStageFlagBits’
  890 |     if ( value & ShaderStageFlagBits::eClusterCullingHUAWEI )

...

```

Indeed `eClusterCullingShaderInvocationsHUAWEI` is available the Vulkan SDK 1.3.243 but not in the vulkan SDK v238 the latest version updated in common file `vulkan/vulkan.hpp`. 

At line `vk_video_samples/vk_video_encoder/../common/include/vulkan/vulkan.hpp:5660` I had to define 
`#define VULKAN_HPP_NO_TO_STRING` to workaround this mismatch use of the include from `common` folder and `/usr/include/vulkan/vulkan_to_string.hpp`

Could we update the Vulkan SDK in common to Vulkan 1.3.243 ?

zlatinski:
This is not the place to report SDK-related issues. Please file an issue at https://github.com/LunarG/VulkanTools/issues.

dabrain34:
[...]

PR #39, opened 2023-05-01
vk_video_encoder: define VULKAN_HPP_NO_TO_STRING

dabrain34:
Mismatch version between 1.3.238 and 1.3.243 leads to issues with vk::to_string


dabrain34:
here is a workaround to fix #38

zlatinski:
Thank you, Stephan!

Issue #41, opened 2023-05-17
Numerous segfaults/asserts on mesa 23.2 with ANV_VIDEO_DECODE


kkartaltepe:
Intel hardware has some initial support for vulkan video decoding but this application appears to fail in numerous ways when attempting to run it on linux+mesa. Most of these appear to be bugs in the sample and not 

1) Configure the program for ANV's queue families and queue count
2) Attempt to run the sample with a 10s clip of Big Bug Bunny
3) Crash.

Trying to get the sample working it seems the fixes/workarounds required were

1) Only `MapMemory`when using `generateColorPatternRgba888` in `vk_video_decoder/libs/VkCodecUtils/VulkanVideoUtils.cpp`  `ImageObject::FillImageWithPattern`, because `vkFillYuv.fillVkImage` also attempts map. Mapping twice is invalid by spec and fails on mesa.
1) Fix dependency on undefined behavior in `Shell::AcquireBackBuffer` where `assert(acquireBuf != nullptr);` attempts to check if the backbuffer queue was empty. So you get more reasonable crashes.
1) disable non-fifo present modes in `vk_video_decoder/libs/VkShell/Shell.cpp` `Shell::ResizeSwapchain`, the code appears to depend on AcquireNextFrame blocking so if other modes are present it will overrun the backbuffer queue and hang or crash. (there is a `vsync` option in the config but this doesnt appear to affect decode presentation).
1) Reduce DPB size checking in` vk_video_decoder/libs/NvVideoParser/include/VulkanH264Decoder.h` `dpb_full()` since the newly allocated reference frame has `state == 0` it is not counted here despite being counted in the dpb's accounting. The next decode step will then assert that the DPB is out of slots and crash when it attempts to allocate another reference frame.

That should cover everything I needed to get the sample decoder running successfully. It should be noted that incorrect handling of ycbcr sampler attachment, synchronization of cmdbufs, image ownership transitions (due to separate video and present queues), and more result in a large amount of validator noise but the frames presented appear more or less correct.

Thanks for writing up the sample as well it was convenient to have something to test the new functionality with.

zlatinski:
Thank you for your feedback and suggestions for app improvements, Kurt!
Would you mind submitting a pull request with all of the changes you'd like to see in the sample?

kkartaltepe:
[...]

PR #42, opened 2023-05-18
Fixes for running on mesa

kkartaltepe:
Fixes from #41 for correctness issues.

You can also see the other changes required to run on mesa in https://github.com/kkartaltepe/vk_video_samples/tree/minimal-runnable that cover the other issues mentioned. But note minimal branch does not render correctly, as the dpb workaround is not correct.

### DIFF
--- a/vk_video_decoder/libs/VkShell/Shell.cpp
+++ b/vk_video_decoder/libs/VkShell/Shell.cpp
@@ -304,9 +304,9 @@ void Shell::AcquireBackBuffer(bool trainFrame) {
     // acquire just once when not presenting
     if (m_settings.noPresent && GetCurrentBackBuffer().GetAcquireSemaphore() != VK_NULL_HANDLE) return;
 
-    AcquireBuffer* acquireBuf = m_ctx.acquireBuffers.front();
+    assert(!m_ctx.acquireBuffers.empty());
 
-    assert(acquireBuf != nullptr);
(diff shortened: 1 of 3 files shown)


zlatinski:
Thank you, Kurt - we'll look at your proposed change very soon.

vk_video_decoder/libs/VkShell/Shell.cpp
```diff
@@ -304,9 +304,9 @@ void Shell::AcquireBackBuffer(bool trainFrame) {
     // acquire just once when not presenting
     if (m_settings.noPresent && GetCurrentBackBuffer().GetAcquireSemaphore() != VK_NULL_HANDLE) return;
 
-    AcquireBuffer* acquireBuf = m_ctx.acquireBuffers.front();
+    assert(!m_ctx.acquireBuffers.empty());
```
zlatinski, line 307: I don't understand why do we need this change?

The rest of the changes make sense and I'll push alternative changes for those.

kkartaltepe: `assert(acquireBuf != nullptr);` depends on `front()` returning a nullptr when empty, but this is undefined behavior and in practice it does not return a nullptr with libstdc++.

This replaces that assert with an assert that acquireBuffers is `!empty()`

The diff maybe looks a bit odd since the assert is moved before the front() call.

zlatinski:
I have already applied these changes in a different form.

PR #50, opened 2023-06-29
Fix double vkUnmap calls

lolzballs:
Appears to be missed in f289815. The memory should be mapped and unmapped by VulkanDeviceMemoryImpl.

Related to #41.


zlatinski (approved):
Thank you, Ben. LGTM.

zlatinski:
Hi Ben, Can you please rebase to the ToT?

PR #62, opened 2023-12-16
Implement `--help` for decoder app CLI and additional CLI arg validation

theHamsta:
This performs three changes to the decoder app CLI parsing
- put the arguments in a data structure to make them available for a `--help` command
- add some additional validation for CLI parameters (missing values, values that start with `-` since likely a mistake). Please let me know if the check for arguments not starting with `-` https://github.com/nvpro-samples/vk_video_samples/pull/62/commits/57e7addf8764ff917bb036288f7787118a89a317 adds too much code and should be omitted. Without it the actual parsing code is quite short https://github.com/nvpro-samples/vk_video_samples/pull/62/commits/be1317ca494d41ea3a8fa0061d2a7844aa406498. https://docs.rs/clap/latest/clap/ performs such a check which had prevented me often to run a program with unintended CLI flags
- Remove flags `noTick` as it seems to be unused. Let me know if they are indented to be used in future.

With this refactor a bug is fixed that prevented users to select `-vv` as `strstr` would also match the `-v` option.

PS: with the recent changes to master I'm now also seeing problems with the decoder with Turing.

### DIFF
--- a/common/libs/VkCodecUtils/ProgramConfig.h
+++ b/common/libs/VkCodecUtils/ProgramConfig.h
@@ -24,9 +24,20 @@
 #include <fstream>
 #include <string>
 #include <vector>
+#include <functional>
+#include <algorithm>
+#include <iomanip> 
+#include <sstream>
 #include "vulkan_interfaces.h"
 
 struct ProgramConfig {
+     struct ArgSpec {
+       const char *flag;
+       const char *short_flag;
+       int numArgs;
+       const char *help;
+       std::function<bool(const char **, const std::vector<ArgSpec> &)> lambda;
+     };
 
     ProgramConfig(const char* programName) {
         appName = programName;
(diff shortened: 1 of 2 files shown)


zlatinski:
Thank you, Stephan! The changes look good, but the change needs a rebase. Can you please do that? Also, we need to make sure this works on Windows.

common/libs/VkCodecUtils/ProgramConfig.h
```diff
@@ -66,88 +76,215 @@ struct ProgramConfig {
...
+                [this](const char **args, const ProgramArgs &a) {
+                    vsync = false;
+                    return true;
+                }},
+            {"--initialWidth", "-w", 1, "Initial width of the video",
+                [this](const char **args, const ProgramArgs &a) {
+                    initialWidth = std::atoi(args[0]);
+                    return true;
+                }},
+            {"--initialHeight", "-h", 1, "Initial height of the video",
```
theHamsta, line 144: before, I changed `-h` from `--initialHeight` to `-h` for `--help` since i was always trying to use `vk-decode-app -h` to list the options. I changed this back following Tymurs feedback.

zlatinski (approved):
Thanks for the change, Stephan! LGTM.

PR #69, opened 2024-02-23
Encode: support separate transfer queue and generic fixes

lolzballs:
These changes were made while trying to get the encoder app working on AMD. The main change here is to support the case where video queues do not support transfer operations. The latter two generic fixes are not really functional, just to help encode more compliant streams.

@zlatinski


zlatinski:
Looks good - thank you for the fix, Ben! I have a follow-up change that changes the way the app checks for video queue support for xfer operations. I've pushed your changes manually to main, so there's no need for these changes to be merged.

zlatinski:
I've pushed your changes manually to main, so there's no need for these changes to be merged.

PR #70, opened 2024-03-13
Fix windows build and correctly wait for the frame to finish decoding before outputting to file.

vasn1k:
Fixes a warning that turns into an error on MSVCs.

Fixes an issue with waiting for fences when the first wait times out.


zlatinski (approved):
Looks good - thank you, Vassili!

PR #73, opened 2024-04-01
decoder: only create compute queue when requested

lolzballs:
If a compute queue isn't requested at InitPhysicalDevice() time, the queue family for compute won't be retrieved. Then if createComputeQueue is set in CreateVulkanDevice, vkGetDeviceQueue will be called with queueFamilyIndex = -1.


zlatinski (approved):
LGTM, thank you for the fix, Ben!

PR #74, opened 2024-04-07
AV1 decode: fix reference frame order hint array

fhvwy:
The code here is intended to implement this logic from the AV1 specification section 5.9.2:

```
for ( i = 0; i < REFS_PER_FRAME; i++ ) {
    refFrame = LAST_FRAME + i
    hint = RefOrderHint[ ref_frame_idx[ i ] ]
    OrderHints[ refFrame ] = hint
}
```

Fix the indexing so that the OrderHints array is set exactly as the specification states rather than offset by one element.  (The offset was propagating into incorrect values for the SavedOrderHints element for reference frames.)


zlatinski:
Thank you, Mark! We agree with your findings and will be working on fixing this in the CTS and the drivers before fixing the sample app.

PR #89, opened 2024-09-27
ci: add a first build and release CI

dabrain34:
I propose to have a CI to validate the future PRs and avoid regressions.

For now there is no format check but it should be added in the future.


zlatinski (approved):
Thank you, Stephane! LGTM. Merging.

PR #90, opened 2024-10-01
CRC support and Y4M support

vasn1k:
Allows the sample app to produce CRCs per frame and for the entire file for decode. The following commandline parameters have been added:
 --crc
 --crcframe
 --crcinit
 --y4m

Fixed --c to actually accept a max frame count instead of silently accepting a bolean as an inteteger..

### DIFF
--- a/common/libs/VkCodecUtils/ProgramConfig.h
+++ b/common/libs/VkCodecUtils/ProgramConfig.h
@@ -75,6 +75,11 @@ struct ProgramConfig {
         enableHwLoadBalancing = false;
         selectVideoWithComputeQueue = false;
         enableVideoEncoder = false;
+        crcOutput = nullptr;
+        outputy4m = false;
+        outputcrcPerFrame = false;
+        outputcrc = false;
+        crcOutputFile = nullptr;
     }
 
     using ProgramArgs = std::vector<ArgSpec>;
@@ -233,7 +238,7 @@ struct ProgramConfig {
             {"--maxFrameCount", "-c", 1,
                 "Limit number of frames to be processed",
                 [this](const char **args, const ProgramArgs &a) {
-                    maxFrameCount = true;
+                    maxFrameCount = std::atoi(args[0]);
                     return true;
                 }},
             {"--queueid", nullptr, 1, "Index of the decoder queue to be used",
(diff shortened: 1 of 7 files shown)


common/libs/VkCodecUtils/VulkanDescriptorSetLayout.h
```diff
@@ -271,13 +271,16 @@ class VulkanDescriptorSetLayout {
             switch (pDescriptorWrite->descriptorType) {
             case VK_DESCRIPTOR_TYPE_SAMPLER:
                 descriptorSize = m_descriptorBufferProperties.samplerDescriptorSize;
-                // fall through
+                __fallthrough;
```
zlatinski: __fallthrough does not work on Linux GCC. Please leave the original change. 
// fall through

common/libs/VkCodecUtils/ProgramConfig.h
```diff
@@ -269,6 +273,42 @@ struct ProgramConfig {
...
+                [this](const char **args, const ProgramArgs &a) {
+                    outputy4m = true;
+                    return true;
+                }},
+            {"--crc", nullptr, 0, "Output a CRC for the entire stream",
+                [this](const char **args, const ProgramArgs &a) {
+                    outputcrc = true;
+                    return true;
+                }},
+            {"--crcframe", nullptr, 0, "Output a CRC for each frame",
```
zlatinski: Can we please call this --crcPerFrame

common/libs/VkCodecUtils/ProgramConfig.h
```diff
@@ -269,6 +273,42 @@ struct ProgramConfig {
...
+                [this](const char **args, const ProgramArgs &a) {
+                    outputcrcframe = true;
+                    return true;
+                }},
+            {"--crcinit", nullptr, 1, "Initial value of the CRC separated by a comma, a set of CRCs can be specified with this commandline parameter",
```
zlatinski, line 297: Can you please write a note on how this option gets to be used?

vk_video_decoder/demos/vk-video-dec/Main.cpp
```diff
@@ -32,6 +32,17 @@ int main(int argc, const char **argv) {
     ProgramConfig programConfig(argv[0]);
     programConfig.ParseArgs(argc, argv);
 
+    // In the regular application usecase the CRC output variables are allocated here and also output as part of main.
+    // In the library case it is up to the caller of the library to allocate the values and initialize them.
+    std::vector<uint32_t> crcAllocation;
```
zlatinski, line 37: I don't think we need a vector to store the CRC values per frame.
Since you have an output CRC file, I would do the following:
Skip the first 16 bytes:
8 bytes for the size of the file and 8 bytes for the global CRS.
Then for each frame, please write sequentially to the file two 16-bit words:
One for the offset of the output and a second one for the CRC. 
At the end of the stream, just move the file pointer back to the beginning and write the total file size and the global CRC.

common/libs/VkCodecUtils/VulkanVideoProcessor.cpp
```diff
@@ -770,3 +754,47 @@ VkResult VulkanVideoProcessor::ParseVideoStreamData(const uint8_t* pData, size_t
 
     return m_vkParser->ParseVideoData(&packet, pnVideoBytes, doPartialParsing);
 }
+
+void VulkanVideoProcessor::WaitForFrameCompletion(VulkanDecodedFrame* pFrame, VkSharedBaseObj<VkImageResource>& imageResource) {
+
+    const uint64_t fenceTimeout = 100 * 1000 * 1000; // 100 mSec
```
zlatinski: We already have a function for that called WaitAndResetFence() Please use it.
I was about to add a function called WaitAndGetStatus() after, but forgot to do it.

zlatinski (changes requested):
Thank you, Vassili for doing that feature change. I left some suggestions.

common/libs/VkCodecUtils/VulkanVideoProcessor.cpp
vasn1k: As you mention,WaitAndResetFence does not do the same thing. So we would need WaitAndGetStatus here.
This isn't really changing anything. We currently have this blob of code in the ConvertFrameToNv12 function. I was just attempting to make the function smaller, and this can go away when you add WaitAndGetStatus().

I'll add a TODO:

vk_video_decoder/demos/vk-video-dec/Main.cpp
vasn1k: This doesn't store the per frame values. It's the set of CRC entries, so you can decide to have multiple CRCs generated with different seeds.

zlatinski: Oho, I see - I misunderstood the change. - sorry!
But I still would like to see the CRC output file with the above format described, please.

common/libs/VkCodecUtils/VulkanVideoProcessor.cpp
zlatinski: You can use WaitAndResetFence() for the first part, or add the WaitAndGetStatus() utility function.

vk_video_decoder/demos/vk-video-dec/Main.cpp
vasn1k: Sure, I was planning to do it in a followup change but I guess now is now.

zlatinski (approved):
Thank you for the changes, Vassili! LGTM. Merging.

PR #91, opened 2024-10-02
Add the missing CRC generator files fix Y4M output.

vasn1k:
These files were missed from the CRC and Y4M support changes.
Additionally Y4M didn't actually end up writing the frame data. - This is now fixed.


zlatinski (approved):
Thank you for the quick fix, Vassili!

PR #95, opened 2024-10-16
Encoder: fallback to layered if separate not present

dabrain34:
As VK_VIDEO_CAPABILITY_SEPARATE_REFERENCE_IMAGES_BIT_KHR is optional
support, the DPB mode should fallback to layered dpb
if the implementation does not support separate dpb.

### DIFF
--- a/vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
@@ -351,8 +351,8 @@ VkResult VkVideoEncoder::InitEncoder(VkSharedBaseObj<EncoderConfig>& encoderConf
     if (encoderConfig->useDpbArray == false &&
         (encoderConfig->videoCapabilities.flags & VK_VIDEO_CAPABILITY_SEPARATE_REFERENCE_IMAGES_BIT_KHR) == 0) {
         std::cout << "Separate DPB was requested, but the implementation does not support it!" << std::endl;
-        assert(!"Separate DPB is not supported");
-        return VK_ERROR_INITIALIZATION_FAILED;
+        std::cout << "Fallback to layered DPB!" << std::endl;
+        encoderConfig->useDpbArray = true;
     }
 
     // Reconfigure the gopStructure structure because the device may not support


vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h
```diff
@@ -443,7 +443,7 @@ struct EncoderConfig : public VkVideoRefCountBase {
     , deviceId(-1)
     , queueId(0)
     , codec(VK_VIDEO_CODEC_OPERATION_NONE_KHR)
-    , useDpbArray(false)
+    , useDpbArray(true)
```
zlatinski: Thank you, Stephan. With the way you've changed the code, we'll end up always using arrays and we don't want that.

There is code at VkVideoEncoder.cpp:351:
    if (encoderConfig->useDpbArray == false &&
        (encoderConfig->videoCapabilities.flags & VK_VIDEO_CAPABILITY_SEPARATE_REFERENCE_IMAGES_BIT_KHR) == 0) {
        std::cout << "Separate DPB was requested, but the implementation does not support it!" << std::endl;
        assert(!"Separate DPB is not supported");
        return VK_ERROR_INITIALIZATION_FAILED;
    }
    
You need an additional change at this place where if the implementation supports VK_VIDEO_CAPABILITY_SEPARATE_REFERENCE_IMAGES_BIT_KHR, then you would set useDpbArray to true.

zlatinski (changes requested):
Hi Stephan, The change makes sense, but it is not complete.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h
dabrain34: On RTX 4060 the two modes are supported by the driver.
So if the implementation supports VK_VIDEO_CAPABILITY_SEPARATE_REFERENCE_IMAGES_BIT_KHR and we will force useDpbArray to false, we will not allow to test the other configuration.  



> VK_VIDEO_CAPABILITY_SEPARATE_REFERENCE_IMAGES_BIT_KHR indicates that the [video picture resources](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#video-picture-resources) associated with the [DPB slots](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#dpb-slot) of a video session can be backed by separate VkImage objects. If this capability flag is not present, then all DPB slots of a video session must be associated with video picture resources backed by the same VkImage object (e.g. using different layers of the same image).

According to the specifications it says that if the implementation does not support `VK_VIDEO_CAPABILITY_SEPARATE_REFERENCE_IMAGES_BIT_KHR` then it **must** support the layered dpb, that's why  I just changed to have the default layered DPB instead of separate DPB.

dabrain34: I just updated the PR to comply with our discussion during the TSG call. The default value should be separate DPB if the implementation supports it and fallback on layered if not.

PR #97, opened 2024-10-17
H264RateControl: do not force the rate control for default

dabrain34:
Intel ANV does not support VBR rate control so keep default rate control selected by the config.

### DIFF
--- a/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
@@ -478,11 +478,7 @@ bool EncoderConfigH264::GetRateControlParameters(VkVideoEncodeRateControlInfoKHR
     pRateControlLayersInfo->frameRateNumerator = frameRateNumerator;
     pRateControlLayersInfo->frameRateDenominator = frameRateDenominator;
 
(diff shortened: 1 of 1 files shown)


vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
```diff
@@ -478,12 +478,6 @@ bool EncoderConfigH264::GetRateControlParameters(VkVideoEncodeRateControlInfoKHR
     pRateControlLayersInfo->frameRateNumerator = frameRateNumerator;
     pRateControlLayersInfo->frameRateDenominator = frameRateDenominator;
 
-    if (rateControlMode == VK_VIDEO_ENCODE_RATE_CONTROL_MODE_DEFAULT_KHR) {
-        pRateControlInfo->rateControlMode = VK_VIDEO_ENCODE_RATE_CONTROL_MODE_VBR_BIT_KHR;
-    } else {
-        pRateControlInfo->rateControlMode = rateControlMode;
```
zlatinski, line 484: So, with this change, we will not support anything but disabled RC?
What happened with:
pRateControlInfo->rateControlMode = rateControlMode;

dabrain34: I have been a bit blunt here :)

zlatinski (approved):
LGTM - Thank you, Stephan!

PR #103, opened 2024-10-28
EncoderConfig: introduce GetMaxBFrameCount

dabrain34:
The default value for B Frame count is 3 but some
implementation does not support B Frames.
Set the value to the max B Frame count supported
by the implementation if it exceeds the value.


dabrain34:
This patch avoids a GPU hang with RADV driver which does not support B-Frames

zlatinski (approved):
The change looks good. Thank you, Stephan!

PR #107, opened 2024-11-05
demos: do not generate twice the HelpersDispatchTable

dabrain34:
The HelpersDispatchTable.* are already generated by the encoder lib.


zlatinski (approved):
Thank you, Stephan, for the fix. It LGTM.

BattleAxeVR:
TBH I am hoping once Vulkan Video APIs stabilize in the Vulkan SDK, that I can skip using the dispatch table + dll hookups in lieu of direct vulkan-1.lib C function calls. 

I suppose that may not happen in a while given how much this is still being actively developed, but long-term it would be nice.

PR #108, opened 2024-11-06
decoder: Fix file output for DISTINCT DPB case

lolzballs:
My thought here is that the transfer operation doesn't have to be done in the distinct case (since its allocated as linear already), so only set `m_useTransferOperation = true` in the coincide mode. Then the `VkVideoFrameBuffer` needs to provide the decode output image views for `IMAGE_VIEW_TYPE_LINEAR`.

Without this change, outputting to a file on a DISTINCT DPB implementation would have caused a crash, since the copy operation would have been attempted with the `dstImage` being NULL.

This still relies on the assumption that a distinct output for decoder images supports linear tiling.

@zlatinski: I'm not sure if this is the proper fix, as I don't fully understand what the image specs and such are supposed to do.


zlatinski (approved):
The change makes sense and LGTM. Thank you, Ben!

PR #111, opened 2024-11-26
Logger: add a log system instead of std::cout

dabrain34:
Add a logger with level from ERROR to DEBUG

### DIFF
--- a/common/include/Logger.h
+++ b/common/include/Logger.h
@@ -1,0 +1,96 @@
+#ifndef LOGGER_H_
+#define LOGGER_H_
+
+#include <iostream>
+#include <fstream>
+#include <string>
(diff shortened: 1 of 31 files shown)


dabrain34:
@zlatinski is there any issue with this PR ? 

The idea here is to enable from the command line the log if necessary with --logLevel 

There are 2 kinds of logs:

 - LOG_S_DEBUG acting as cout/cerr
 - LOG_DEBUG acting as printf.

Please consider this patch as it can be very useful and avoid the use of conditional such as:

```
if(verbose)
...
```

dabrain34:
Closed and moved to https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/5

PR #112, opened 2024-11-26
EncoderConfig: add input file get frame count

dabrain34:
In case user does not specify --numFrames, the gopFrameCount was zero and causing a crash.
If user does not specify the numFrames, the number of frames in the input file should be used.

### DIFF
--- a/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.cpp
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.cpp
@@ -133,6 +133,7 @@ int EncoderConfig::ParseArguments(int argc, char *argv[])
     int argcount = 0;
     std::vector<char*> arglist;
     std::vector<std::string> args(argv, argv + argc);
+    uint32_t frameCount = 0;
 
     appName = args[0];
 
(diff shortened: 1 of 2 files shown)


dabrain34:
@zlatinski This patch was missing in the rebase

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.cpp
```diff
@@ -524,6 +525,19 @@ int EncoderConfig::ParseArguments(int argc, char *argv[])
         return -1;
     }
 
+    frameCount = inputFileHandler.GetFrameCount(input.width, input.height, input.bpp, input.chromaSubsampling);
```
zlatinski, line 528: What if the user wanted to specify a frame count smaller than the full video length?

dabrain34: if the user does not set `numFrames then it will be equal to `frameCount`, number of frames detected in the source. Otherwise it will be `numFrames` set by the user which has to be lower or equal to `frameCount`. 

This patch avoid a gopFrameCount equal to 0 leading to a division by zero

zlatinski: OK, thank you, Stepahan!

zlatinski (approved):
Looks good - thank you!

PR #113, opened 2024-11-27
EncoderConfigAV1: Fix GetMaxBFrameCount

dabrain34:
To support driver which does not support B Frame,
GetMaxBFrameCount has been introduced in
https://github.com/nvpro-samples/vk_video_samples/pull/103 but AV1 was not supported yet.


dabrain34:
@zlatinski this one is necessary for CTS

zlatinski (approved):
LGTM.

PR #115, opened 2024-11-28
EncoderConfigAV1: add missing override

dabrain34:
GetMaxBFrameCount needs to be overriden in the case of AV1Config


dabrain34:
@zlatinski ping ^

PR #116, opened 2024-12-02
Fixed inconsistency in sve parser code

t-boiko:
Error in a rare case if the datasize less than lanes number


dabrain34:
Do you have a sample or a use case to demonstrate the issue ?

zlatinski (approved):
LGTM.

PR #117, opened 2024-12-02
Fixed potential rare corner cases

t-boiko:
a potential out-of-bondary access when the next start code isn't found in a SVE - for datasize < 2, for others - when the loop ends without finding the code and i == datasize (next iteration outside loop)


zlatinski (approved):
LGTM.

PR #118, opened 2024-12-02
eliminated unneeded offset

t-boiko:
We have already rounded down the "datasize32" value, and introduced check for i >= datasize in the block. No need for the additional extra offset


t-boiko:
Do not merge, I will double check that

PR #119, opened 2024-12-02
Revert unneeded check

t-boiko:
The checks are redundant because the maximum possible value after the loop ends is `datasizeN - N`, loop step is `N` (I introduced them in a previous commit). Only the SVE-based code's change from the previous commit is required (left it)


zlatinski (approved):
LGTM

PR #123, opened 2025-01-26
Added usage and content hints parameters to encoder

t-boiko:
Added ability to set in VkVideoEncodeUsageInfoKHR:
- VkVideoEncodeUsageFlagsKHR;
- VkVideoEncodeContentFlagBitsKHR.
in addition to VkVideoEncodeTuningModeKHR.


dabrain34:
I would recommend you to open the PR on https://github.com/KhronosGroup/Vulkan-Video-Samples which owns github actions and might be more reactive.

Otherwise LGTM.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.cpp
```diff
@@ -54,6 +54,10 @@ void printHelp(VkVideoCodecOperationFlagBitsKHR codec)
     --lastFrameType                 <integer> : Last frame type \n\
     --closedGop                     Close the Gop, default open\n\
     --qualityLevel                  <integer> : Select quality level \n\
+    --usageHints                    <integer> or <string> : Select encode usage hints \n\
```
zlatinski: I prefer not to use magic numbers as parameters. Please use the string names of the parameters, like : default, transcoding, streaming, etc. Make sure lowercase and uppercase are working.
Same for the tuning mode. I've missed that one, but please add name strings for it, too.

zlatinski (changes requested):
A few minor changes are required, please.

t-boiko:
Done. All these arguments are strings and are casted to the lowercase strings.

zlatinski (approved):
LGTM. Thank you, Tymur!

PR #124, opened 2025-03-03
Added 1-to-n transcoding

t-boiko:
A brief explanation

Transcoding is an extension of the decoder - it is based on both the decoder and the encoder
- transcoding from one codec to another is supported (you can specify a target bitrate, a vbv buffer relative size, vbr or cbr as well);
- transcoding to multiple different resolutions.

Transcoder specific changes are put together with the decoder's and the encoder's code because from maintenance and support points of view it would be time consuming to track and update everything from the decoder and the encoder. My goal was to minimize the code duplication and efforts for the code maintenance and support (however the transcoder has it's separate own project folder and Main).

The transcoder can be built from it's own folder using CMake in the same way as for the decoder or the encoder (see Build.md in the transcoder's folder). 

cmdline example:
./demos/vk-video-trans-test -i input_file_to_reencode  --codec hevc --qualityLevel 3 --profile 0 --bitrate 5  --vbvbuf-ratio 2 --numberResizedOutputs 3 1280x720 800x600 352x288

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -375,6 +375,7 @@ endif()
 
     if(BUILD_DEMOS AND NOT DEFINED DEQP_TARGET)
         add_subdirectory(vk_video_decoder/demos)
+        add_subdirectory(vk_video_transcoder/demos)
     endif()
 endif()
 ############ VULKAN_VIDEO_DECODER_LIB ######################################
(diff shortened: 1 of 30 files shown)


zlatinski:
@t-boiko, please move this to a new dev-branch.

PR #126, opened 2025-03-07
AV1 profile and tier selection fix.

krajunv:
Include compressed rate and different tiers to validate level support.
Update the sequence header with level and tier parameters with supported values.


zlatinski (approved):
LGTM. Thank you!

PR #127, opened 2025-03-07
Set the bitstream buffer size to match the input image size.

krajunv:
For higher image resolutions and bitrates, the fixed 2MB buffer size is inadequate. To handle the worst-case scenario, ensure the bitstream size is the same as the input image size.


zlatinski (approved):
LGTM. Thank you

PR #128, opened 2025-03-11
Return error if the specified qualityLevel is greater than the maximu…

krajunv:
The number of quality levels supported by the implementation can vary for each codec and may also depend on other parameters.

Therefore, if a client specifies a qualityLevel that is not supported by the implementation, this change fails encode initialization. 

Even if the client attempts to use out-of-range values by disabling this check, the validation layer must catch this error if it is enabled. The implementation may either clamp the value to a valid range or return an error.


zlatinski (approved):
LGTM. Thank you!

PR #130, opened 2025-03-18
Instead of defining the destructor inline, provide its implementation…

krajunv:
… in a separate source file to make GCC 13 happy in release mode.

Also fix other compilation errors in release mode.

### DIFF
--- a/common/libs/VkCodecUtils/VkImageResource.cpp
+++ b/common/libs/VkCodecUtils/VkImageResource.cpp
@@ -88,6 +88,11 @@ VkImageResource::VkImageResource(const VulkanDeviceContext* vkDevCtx,
     }
 }
 
+VkImageResource::~VkImageResource()
+{
+    Destroy();
+}
+
 VkResult VkImageResource::Create(const VulkanDeviceContext* vkDevCtx,
                                  const VkImageCreateInfo* pImageCreateInfo,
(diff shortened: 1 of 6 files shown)


0001-fix-compilation-errors-in-release-mode.patch
```diff
@@ -0,0 +1,102 @@
...
+ VkResult VkImageResource::Create(const VulkanDeviceContext* vkDevCtx,
+                                  const VkImageCreateInfo* pImageCreateInfo,
+                                  VkMemoryPropertyFlags memoryPropertyFlags,
+diff --git a/common/libs/VkCodecUtils/VkImageResource.h b/common/libs/VkCodecUtils/VkImageResource.h
+index 314a2f0..0c4c0ac 100644
+--- a/common/libs/VkCodecUtils/VkImageResource.h
++++ b/common/libs/VkCodecUtils/VkImageResource.h
+@@ -113,7 +113,7 @@ private:
+ 
+     void Destroy();
```
zlatinski: I guess, we don't need this patch submitted here.
Otherwise, the change looks good to me.

krajunv: Removed the file.

zlatinski (approved):
LGTM. Thank you, Konda!

PR #133, opened 2025-05-01
github actions: Build Windows only with FFMPEG

zlatinski:


dabrain34:
Your commit does not explain why you removed this job which helped to check if there is no regression in a system without ffmpeg

zlatinski:
> Your commit does not explain why you removed this job which helped to check if there is no regression in a system without ffmpeg

Hi Stephan, for Nvidia, using  the compute filter, and a compute queue, in general, is essential. We don't need a build for  "no  shaderc", because nvpro will not support a configuration without shaderc.

dabrain34:
> > Your commit does not explain why you removed this job which helped to check if there is no regression in a system without ffmpeg
> 
> Hi Stephan, for Nvidia, using the compute filter, and a compute queue, in general, is essential. We don't need a build for "no shaderc", because nvpro will not support a configuration without shaderc.

Not sure to understand how it is related to ffmpeg. I disabled shaderc in Vulkan-Video-Samples to avoid the encoder build issue with CTS but I dont think its related to ffmpeg.

PR #137, opened 2025-05-02
Hw laod balancing

zlatinski:
Switch to VkSubmitInfo2KHR and use TL semaphores

    Switch the submit API to VK_KHR_synchronization2 using
    VkSubmitInfo2KHR.
    Use more precise stage masks for synchronization.
    Switch the decoder to use timeline semaphores. This technique reduces
    resource use and improves the synchronization.

    There are now two timelines used in the decoder's FB:
    1. decode timeline (producer) that follows the decode order
    2. display timeline (consumer) that follows the display order

    The decode timeline semaphore also supports a filter synchronization
    using the same semaphore, but it increments the sem. value by 2.

    Add 20 ms sleep for debuging.

common and decode: HW LoadBalancing fixes

    HW LoadBalancing value starts at zero
    Context Encode default value

encode: add support for HW load-balancing


zlatinski:
Already merged.

PR #138, opened 2025-05-06
Report invalid cmdline arguments in h264 and hevc encoder

t-boiko:
Exit and warn If command line string consists of invalid argument

### DIFF
--- a/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h
@@ -936,7 +936,17 @@ public:
 
     int ParseArguments(int argc, char *argv[]);
 
-    virtual int DoParseArguments(int argc, char *argv[]) { return 0; };
+    virtual int DoParseArguments(int argc, char *argv[]) {
+        if (argc > 0) {
+            std::cout << "Invalid paramters: ";
+            for (int i = 0; i < argc; i++) {
+                std::cout << argv[i] << " ";
+            }
+            std::cout << std::endl;
+            return -1;
+        }
+        return 0;
+    };
 
     virtual VkResult InitializeParameters()
     {


vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h
```diff
@@ -936,7 +936,17 @@ struct EncoderConfig : public VkVideoRefCountBase {
 
     int ParseArguments(int argc, char *argv[]);
 
-    virtual int DoParseArguments(int argc, char *argv[]) { return 0; };
+    virtual int DoParseArguments(int argc, char *argv[]) {
```
zlatinski, line 939: I don't understand this change. EncoderConfig::ParseArguments() always calls DoParseArguments() at the end. Do you rely on argcount to get the remaining bad arguments?

t-boiko: This change causes the program to terminate when an incorrect argument is passed (currently the program does not terminate, for example, you can try passing `--tuningInfo` instead of `--tuningMode`). At the first glance it looks like relying on the remaining arguments' count

zlatinski (approved):
LGTM. Thank you!

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h
krajunv: The `DoParseArguments()` method is responsible for parsing codec-specific arguments that are not handled by `EncoderConfig::ParseArguments()`.

For the H264 and HEVC codecs, the application does not support any additional codec-specific arguments. As a result, the `DoParseArguments()` method is not overridden in their respective classes. If any arguments remain unparsed by `EncoderConfig::ParseArguments()`, the method returns an error; otherwise, it performs no action.

Only the AV1 codec implementation supports additional codec-specific arguments. In this case, the `DoParseArguments()` method is overridden to handle those specific parameters.

PR #139, opened 2025-05-17
Use compute filter copy convert

zlatinski:
encode: remove the CPU input conversion function
    
    Compute filter: Add support for bufffer as input
    Compute filter: Add support for in/out plane mismatch
    Compute filter: Add 10/12-bit shift support
    Compute filter: Add supprt for 4:4:4, 4:2:2 and 4:2:0


zlatinski:
Tymur has tested it to work.

PR #140, opened 2025-05-17
Make the code more generic for different compiler settings

vasn1k:
This fixes some common compiler issues seen with the sample application when used outside of the cmake build environment with more strict settings.


dabrain34:
can it be related and merged with https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/7 ?

vasn1k:
> can it be related and merged with [KhronosGroup/Vulkan-Video-Samples#7](https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/7) ?

I didn't see too much overlap in the changes, so I think they should get merged in parallel. I don't expect merge conflicts.

common/libs/VkCodecUtils/DecoderConfig.h
```diff
@@ -40,6 +41,14 @@ struct DecoderConfig {
        std::function<bool(const char **, const std::vector<ArgSpec> &)> lambda;
      };
 
+    static inline void safe_exit(int status) {
+    #if defined(DECODER_APP_BUILD_AS_LIB) || defined(ENCODER_APP_BUILD_AS_LIB)
+        printf("Application would exit with status %d\n", status);
```
zlatinski: ENCODER_APP_BUILD_AS_LIB is confusing because both the encoder and decoder are built as libraries.

The issue here is that ParseArgs() doesn't return a status value but rather exits directly.

Please remove the ExitWrapper.h and all this code and simply return the error code.

Then
decoderConfig.ParseArgs(argc, argv);
becomes
bool success = decoderConfig.ParseArgs(argc, argv);
if (!success) {
   message and return or whatever.
}

common/libs/VkCodecUtils/VulkanVideoProcessor.cpp
```diff
@@ -174,12 +177,11 @@ VkResult VulkanVideoProcessor::Create(const DecoderConfig& settings, const Vulka
 
 VkVideoProfileInfoKHR VulkanVideoProcessor::GetVkProfile() const
 {
-
-    VkVideoProfileInfoKHR videoProfile({VK_STRUCTURE_TYPE_VIDEO_PROFILE_INFO_KHR, NULL,
+    VkVideoProfileInfoKHR videoProfile = {VK_STRUCTURE_TYPE_VIDEO_PROFILE_INFO_KHR, NULL,
```
zlatinski: Why do we need this change? Are we targeting an ancient compiler?

common/libs/VkCodecUtils/VulkanVideoProcessor.cpp
```diff
@@ -209,10 +211,10 @@ VkFormat VulkanVideoProcessor::GetFrameImageFormat()  const
 
 VkExtent3D VulkanVideoProcessor::GetVideoExtent() const
 {
-    VkExtent3D extent ({ (uint32_t)m_videoStreamDemuxer->GetWidth(),
+    VkExtent3D extent = { (uint32_t)m_videoStreamDemuxer->GetWidth(),
```
zlatinski: Why do we need this change to copy rather than init? Are we targeting an ancient compiler?

common/libs/VkShell/ShellWayland.cpp
```diff
@@ -195,7 +196,7 @@ void ShellWayland::InitConnection() {
         if (!m_shell) throw std::runtime_error("failed to bind shell");
     } catch (const std::exception &e) {
         std::cerr << "Could not initialize Wayland: " << e.what() << std::endl;
-        exit(-1);
+        safe_exit(-1);
```
zlatinski: Why is this exit is a problem? This code is not part of the "library", right?

vk_video_decoder/include/vulkan_video_decoder.h
```diff
@@ -44,9 +44,9 @@
 #include "VkDecoderUtils/VideoStreamDemuxer.h"
 
 // High-level interface of the video decoder
-class VulkanVideoDecoder : public VkVideoQueue<VulkanDecodedFrame> {
+class VulkanVideoDecoderHighLevel : public VkVideoQueue<VulkanDecodedFrame> {
```
zlatinski: Why do we need this change?

vk_video_decoder/include/vulkan_video_decoder.h
```diff
@@ -44,9 +44,9 @@
...
 public:
-    virtual ~VulkanVideoDecoder() {};
+    virtual ~VulkanVideoDecoderHighLevel() {};
```
zlatinski: Why do we need this change?

vk_video_decoder/libs/NvVideoParser/src/VulkanAV1Decoder.cpp
```diff
@@ -2281,7 +2281,11 @@ bool VulkanAV1Decoder::ParseObuTileGroup(const AV1ObuHeader& hdr)
             consumedBytes += tile_size_bytes_minus_1 + 1;
             m_PicData.tileOffsets[m_PicData.khr_info.tileCount] = (uint32_t)m_nalu.start_offset + (uint32_t)consumedBytes;
 
-            tileSize = tile_size_minus_1 + 1;
+            // Add bounds checking and safe conversion
+            if (tile_size_minus_1 > SIZE_MAX - 1) {
+                return false; // Tile size too large
```
zlatinski, line 2286: Code style:
if (tile_size_minus_1 > (SIZE_MAX - 1)) {

vk_video_decoder/libs/NvVideoParser/src/VulkanVideoDecoder.cpp
```diff
@@ -319,33 +319,32 @@ VkDeviceSize VulkanVideoDecoder::swapBitstreamBuffer(VkDeviceSize copyCurrBuffOf
 bool VulkanVideoDecoder::ParseByteStream(const VkParserBitstreamPacket* pck, size_t *pParsedBytes)
 {
 #if defined(__x86_64__) || defined (_M_X64)
-    if (m_NextStartCode == SIMD_ISA::AVX512)
-    {
+    #if defined(__AVX512F__) && defined(__AVX512BW__)
```
zlatinski: Please pull out this change as a separate patch and let Tymur review it. We support multiple platforms and compilers and have to be cautious with these types of changes.

vk_video_decoder/src/vulkan_video_decoder.cpp
```diff
@@ -21,7 +21,7 @@
 #include "VkVideoDecoder/VkVideoDecoder.h"
 #include "VkCodecUtils/VulkanVideoProcessor.h"
 
-class VulkanVideoDecoderImpl : public VulkanVideoDecoder {
+class VulkanVideoDecoderImpl : public VulkanVideoDecoderHighLevel {
 public:
```
zlatinski: Why do we need this change?

vk_video_decoder/src/vulkan_video_decoder.cpp
```diff
@@ -66,10 +66,11 @@ class VulkanVideoDecoderImpl : public VulkanVideoDecoder {
 
     virtual VkExtent3D GetVideoExtent() const
     {
-        VkExtent3D extent ({ (uint32_t)m_vulkanVideoProcessor->GetWidth(),
-                             (uint32_t)m_vulkanVideoProcessor->GetHeight(),
-                             (uint32_t)1
-                           });
+        VkExtent3D extent = {
+            (uint32_t)m_vulkanVideoProcessor->GetWidth(),
```
zlatinski, line 70: Why switch from initialization to assignment? This approach is less efficient.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderDpbH264.cpp
```diff
@@ -278,7 +279,7 @@ int8_t VkEncDpbH264::DpbPictureEnd(const PicInfoH264 *pPicInfo,
             }
             if (m_currDpbIdx >= MAX_DPB_SLOTS) {
                 VK_DPB_DBG_PRINT(("could not allocate a frame buffer\n"));
-                exit(1);
+                safe_exit(1);
```
zlatinski: I don't mind using your exit wrapper to places like this one, but I would like to extend it with the function name, line number, and error text. 

Like:
safe_exit(__function_, __line_, "H264 parser could not allocate a frame buffer\n");

vk_video_encoder/src/vulkan_video_encoder.cpp
```diff
@@ -83,7 +83,7 @@ class VulkanVideoEncoderImpl : public VulkanVideoEncoder {
 VkResult VulkanVideoEncoderImpl::Initialize(VkVideoCodecOperationFlagBitsKHR videoCodecOperation,
                                             int argc, char** argv)
 {
-    VkResult result = EncoderConfig::CreateCodecConfig(argc, argv, m_encoderConfig);
+    VkResult result = EncoderConfig::CreateCodecConfig(argc, (const char**)argv, m_encoderConfig);
     if (VK_SUCCESS != result) {
```
zlatinski, line 87: The remaining modifications from this change LGTM. Thank you for the fixes, Vassili!

zlatinski (changes requested):
Thank you, Vassili, for the changes! I would like you to make some changes to your patch, please.

common/libs/VkCodecUtils/VulkanShaderCompiler.cpp
```diff
@@ -103,17 +103,19 @@ VkShaderModule VulkanShaderCompiler::BuildShaderFromFile(const char *fileName,
                                                          VkShaderStageFlagBits type,
                                                          const VulkanDeviceContext* vkDevCtx)
 {
+#ifdef seekg
```
zlatinski, line 106: What happens if seekg is not defined?

common/libs/VkCodecUtils/VulkanVideoProcessor.h
```diff
@@ -23,6 +23,14 @@
 #include "VkCodecUtils/VkVideoQueue.h"
 #include "VkVideoFrameOutput.h"
 
+// Forward declarations
+class VulkanDeviceContext;
+struct VkMpFormatInfo;
+
+size_t ConvertFrameToNv12(const VulkanDeviceContext* vkDevCtx, int32_t frameWidth, int32_t frameHeight,
```
zlatinski: I don't understand why we need this change. VulkanVideoProcessor.cpp doesn't need ConvertFrameToNv12. It is only needed and used by VkVideoFrameToFile.cpp and this is the file provider for the decoder library.

vk_video_decoder/libs/VkDecoderUtils/VideoStreamDemuxer.cpp
```diff
@@ -24,7 +24,9 @@ bool VideoStreamDemuxer::CheckFile(const char* szInFilePath)
     if (fpIn.fail()) {
         std::ostringstream err;
         err << "Unable to open input file: " << szInFilePath << std::endl;
+#ifdef __cpp_exceptions
         throw std::invalid_argument(err.str());
```
zlatinski, line 29: Maybe we need a safe_exit() here if exceptions are not supported?

zlatinski (changes requested):
Thank you, Vassili! Please look at my below comments. I still need to review your Vulkan chain changes.

common/libs/VkCodecUtils/VkVideoFrameToFile.cpp
```diff
@@ -272,7 +272,7 @@ class VkVideoFrameToFileImpl : public VkVideoFrameOutput {
                       << std::endl;
             fileNameWithModExt = fileName + std::string(".y4m");
             fileName = fileNameWithModExt.c_str();
-        } else if (!hasExtension(fileName, ".yuv")) {
+        } else if (y4mFormat == false && !hasExtension(fileName, ".yuv")) {
             std::cout << std::endl << "Raw yuv output format is requested, ";
```
zlatinski, line 211: Clang code style, please.
} else if ((y4mFormat == false) && !hasExtension(fileName, ".yuv")) {

common/libs/VkCodecUtils/VkVideoFrameToFile.cpp
```diff
@@ -239,6 +241,12 @@ class VkVideoFrameToFileImpl : public VkVideoFrameOutput {
             }
         }
 
+        if (m_crcOutput != nullptr) {
+            for (size_t i = 0; i < m_crcInitValue.size(); i += 1) {
+                getCRC(&m_crcOutput[i], pOutputBuffer, usedBufferSize, Crc32Table);
```
zlatinski, line 173: I don't understand why your client code doesn't call getCRC() directly but needs to pass a pointer to output CRC? Furthermore, your CRC pointer doesn't have a size, and the copy doesn't look safe to me.

common/libs/VkCodecUtils/VulkanShaderCompiler.cpp
vasn1k: if seekg is undefined, the code will just fail to build any shaders from file. This function also seems untested - see my other change to it. Hint: tellg was being invoked before any seeking was done.

vk_video_decoder/src/vulkan_video_decoder.cpp
vasn1k: There is a namespace clash with VulkanVideoDecoder in the codebase.

vk_video_decoder/libs/NvVideoParser/src/VulkanVideoDecoder.cpp
vasn1k: Would it be possible for Tymur to review it here directly? Splitting the patch out would just complicate this merge as it can't happen without these changes. 
This change was made to allow compiling without SIMD optimizations enabled, as they are less important in a test environment setting. Before the compilation would fail.

vk_video_decoder/include/vulkan_video_decoder.h
vasn1k: Namespace clash on VulkanVideoDecoder with the lib and the non-lib code.

common/libs/VkShell/ShellWayland.cpp
vasn1k: I did a search replace to prevent issues and divergence in the future. I'll incorporate your feedback related to line number and some explanation.

common/libs/VkCodecUtils/VulkanVideoProcessor.cpp
vasn1k: I'm also slightly confused what the original code was attempting to do, VkVideoProfileInfoKHR is a structure not a class and does not have a constructor. I removed the copy as you suggested.

common/libs/VkCodecUtils/VulkanVideoProcessor.cpp
vasn1k: VulkanVideoProcessor.cpp(212): error C2100: illegal indirection
This is with msbuild version 16.11.2

I'm also slightly confused what the original code was attempting to do, VkExtent3D is a structure not a class and does not have a constructor. I removed the copy as you suggested.

common/libs/VkCodecUtils/VkVideoFrameToFile.cpp
vasn1k: > I don't understand why your client code doesn't call getCRC() directly but needs to pass a pointer to output CRC? Furthermore, your CRC pointer doesn't have a size, and the copy doesn't look safe to me.

It's up to the caller to allocate as many space in crcOutput as there are entries in crcInitValues. The m_crcInitValue was being passed into this function and this is why I chose to pass in m_crcOutput as well.
The reasoning for needing to do this here was: VkVideoFrameToFile doesn't expose the pOutputBuffer, outside of the OutputFrame function and I wanted to avoid having a callback function just to enable this.

common/libs/VkCodecUtils/VulkanVideoProcessor.h
vasn1k: This was needed as the GCC compiler warns that ConvertFrameToNv12 has no forward declaration. Since the function is used in VideoFrameToFile, I thought it should be exposed in the header file. I realize now that we have a duplicate of the ConvertFrameToNv12 function in VideoFrameToFile.cpp. I've put the forward decl in the VulkanVideoProcessor.h instead.

common/libs/VkCodecUtils/DecoderConfig.h
vasn1k: Made a new function for this that the implementing library can decide to overload.

vk_video_decoder/libs/NvVideoParser/src/VulkanAV1Decoder.cpp
```diff
@@ -29,6 +29,10 @@
 
 #define ARRAY_SIZE(arr) (sizeof(arr) / sizeof(arr[0]))
 
+// Forward declarations
```
srinathkr-nv: Instead of these forward declarations, I think it will be cleaner and easier to simply add `static` in front of the actual function definitions below.

vk_video_decoder/libs/NvVideoParser/src/VulkanAV1GlobalMotionDec.cpp
```diff
@@ -82,6 +82,9 @@
 #define WARP_PARAM_REDUCE_BITS 6
 #define WARPEDMODEL_PREC_BITS 16
 
+// Function declaration
```
srinathkr-nv: Instead of this forward declaration, the next line can be just `static int get_msb(unsigned int n)`

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
```diff
@@ -1644,7 +1642,7 @@ VkResult VkVideoEncoder::SubmitVideoCodingCmds(VkSharedBaseObj<VkVideoEncodeFram
     return result;
 }
 
-VkResult VkVideoEncoder::PushOrderedFrames()
+VkResult VkVideoEncoder::PushOrderedFrames()
```
srinathkr-nv: Nit: this looks like a spurious whitespace edit.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderDpbH264.h
```diff
@@ -133,7 +133,7 @@ class VkEncDpbH264
     const StdVideoEncodeH264PictureInfo *GetCurrentDpbEntry(void)
     {
         assert((m_currDpbIdx < m_max_dpb_size) || (m_currDpbIdx == MAX_DPB_SLOTS));
-        return &m_DPB[m_currDpbIdx].picInfo;
+        return &m_DPB[(int)m_currDpbIdx].picInfo;
```
srinathkr-nv, line 136: What is the compiler warning/error which makes this cast necessary?

common/libs/VkCodecUtils/Helpers.h
```diff
@@ -338,17 +338,16 @@ inline VkBaseInStructure* ChainNextVkStruct(NodeType& node, ChainedNodeType& nex
...
-    while (pNode->pNext != nullptr) {
-         pNode = (VkBaseInStructure*)pNode->pNext;
-     }
-     pNode->pNext = (VkBaseInStructure*)&nextChainedNode;
-     // make sure the nextChainedNode is of type VkBaseInStructure
-     assert(nextChainedNode.sType > 0);
-     assert(nextChainedNode.pNext == nullptr);
-     return (VkBaseInStructure*)nextChainedNode.pNext;
- }
+    VkBaseInStructure* pNode = (VkBaseInStructure*)(&node);
```
srinathkr-nv, line 343: I'm glad that this turned out to be a simple change (simple both in terms or code and readability), but please pull out this change and the other changes involving the use of `vk::ChainNextVkStruct` into a separate commit within this PR.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderDpbH264.h
vasn1k: The issue was converting a char into an index: "array subscript is of type 'char'"
m_currDpbIdx is typed as an int8_t, which is typed as: typedef signed char int8_t.

common/libs/VkCodecUtils/Helpers.h
vasn1k: It'd become a mess with spot-reverts. Sorry, I'll keep that in mind next time. But for now please search for ChainNextVkStruct that will give you a good indication what was changed.

vk_video_decoder/libs/NvVideoParser/src/VulkanVideoDecoder.cpp
vasn1k: Updated the MR to have this as a separate commit.

vk_video_decoder/libs/VkDecoderUtils/ElementaryStream.cpp
```diff
@@ -136,7 +135,12 @@ class ElementaryStream : public VideoStreamDemuxer {
 
     virtual ~ElementaryStream() {
 #ifdef USE_SIMPLE_MALLOC
+        if (m_videoCodecType == VK_VIDEO_CODEC_OPERATION_DECODE_AV1_BIT_KHR) {
```
zlatinski, line 138: LGTM

common/libs/VkCodecUtils/VkVideoFrameToFile.cpp
```diff
@@ -218,7 +218,9 @@ class VkVideoFrameToFileImpl : public VkVideoFrameOutput {
         if (fileName != nullptr) {
             m_outputFile = fopen(fileName, "wb");
             if (m_outputFile) {
+#if !defined(REDUCE_VK_VIDEO_STDOUT_SPEW)
```
zlatinski, line 221: Please just commend this line out. Now we also have a logger in the Khronos repository, and I would rather not have so many ways to disable logging.

common/libs/VkCodecUtils/VulkanFrame.cpp
```diff
@@ -291,6 +291,7 @@ bool VulkanFrame<FrameDataType>::OnFrame( int32_t renderIndex,
     const bool gfxRendererIsEnabled = (m_videoRenderer != nullptr);
     m_frameCount++;
 
+#if !defined(REDUCE_VK_VIDEO_STDOUT_SPEW)
```
zlatinski, line 294: For all these locations, please add a local runtime config like m_printInfo

common/libs/VkCodecUtils/VulkanVideoProcessor.cpp
zlatinski: A structure in C++ is a class with default parameters in public scope. When a constructor is not explicitly provided, the compiler will provide a default one. Please don't change codes that don't need to be fixing Again, initialization constructors are more efficient than assignments, and this code makes things worse and not better.

common/libs/VkCodecUtils/DecoderConfig.h
```diff
@@ -109,7 +118,7 @@ struct DecoderConfig {
             {"--help", nullptr, 0, "Show this help",
                 [argv](const char **, const ProgramArgs &a) {
                     int rtn = showHelp(argv, a);
-                    exit(EXIT_SUCCESS);
+                    safe_exit(EXIT_SUCCESS);
```
zlatinski: I still see the original code here. Where is the change to a return value? Am I looking at the fixes here or the original change?

common/libs/VkCodecUtils/DecoderConfig.h
```diff
@@ -118,7 +110,7 @@ struct DecoderConfig {
             {"--help", nullptr, 0, "Show this help",
                 [argv](const char **, const ProgramArgs &a) {
                     int rtn = showHelp(argv, a);
-                    safe_exit(EXIT_SUCCESS);
+                    SAFE_EXIT(EXIT_SUCCESS, "Help requested - exiting successfully");
```
zlatinski, line 113: Oho, this is a separate change with the fixes. I thought you'd fixed the original commit. Let me look again, please.

common/libs/VkCodecUtils/VkVideoFrameToFile.cpp
vasn1k: I didn't want to comment that line out because it would affect how the sample app works currently. My goal is to make these changes not impact the apps logic or output at all.

common/libs/VkCodecUtils/VulkanFrame.cpp
vasn1k: If changes are incoming that remove the need for this change, then lets not impact the runtime.

PR #142, opened 2025-06-02
Fix parser build on Linux aarch64

srinathkr-nv:
This is a regression from abbd907786ff67012eedab5ded328a346e0f73fe.


zlatinski (approved):
LGTM. Sorry for the regression, and thanks for the fix, Srinath!

PR #144, opened 2025-06-05
determine levelIdc first to estimate dpbsize

krajunv:
The current derivation DPB size uses default leveIdc value irrespective of actual levelIdc derived.
To estimate required DPB size, determine leveIdc first.  Chose MaxDPBMbs from Table A-1 from H264 spec using levelIdc and derive DPB size


zlatinski (approved):
LGTM. Thank you for the fix, Raju!

PR #145, opened 2025-06-09
Vassili's compilation fixes

zlatinski:
common: Code compilation fixes
common: SAFE_EXIT macro to print and exit
common: Use the CRC generator instead of embedding it
encoder: Deal with the Vulkan chained stuctures
Fix frame to file from adding .yuv to filenames with .y4m already

@vasn1k @srinathkr-nv please review.


common/libs/VkCodecUtils/ExitWrapper.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include <cstdio>
+#include <cstdlib>
+
+inline void safe_exit(int status, const char* function_name, int line_number, const char* description) {
+    printf("Exiting from %s:%d - %s (status: %d)\n", function_name, line_number, description, status);
+    exit(status);
```
vasn1k: Using an additional file was done so implementations of this code could leave it out and easily define their own safe_exit function. Defining safe_exit in the header, requires additional MACRO's to go around the exit call, preventing it from being compiled in. So there we would need the BUILD_DECODER_APP_AS_LIB check which will switch between exit and printf.

common/libs/VkCodecUtils/VkVideoFrameToFile.cpp
```diff
@@ -272,7 +199,7 @@ class VkVideoFrameToFileImpl : public VkVideoFrameOutput {
                       << std::endl;
             fileNameWithModExt = fileName + std::string(".y4m");
             fileName = fileNameWithModExt.c_str();
-        } else if (!hasExtension(fileName, ".yuv")) {
+        } else if (y4mFormat == false && !hasExtension(fileName, ".yuv")) {
```
vasn1k: } else if ((y4mFormat == false) && !hasExtension(fileName, ".yuv")) {

zlatinski: Yes, thank you! Fixed with the update.

zlatinski:
> cpudetect changes are missing.

Yes, I've included that one two.

zlatinski:
> So we don't forget, below changes have been dropped but are necessary:

> m_crcOutput is missing, important.

This one needs a more robust change using a local vector that the client would request to copy from. 

> REDUCE_VK_VIDEO_STDOUT_SPEW is missing.

Can we rename this one to VK_VIDEO_NO_STDOUT_INFO?

vasn1k:
> (m_crcOutput) This one needs a more robust change using a local vector that the client would request to copy from.

I think this boils down to preference, my preference was caller allocation and management. Your voiced preference is lib ownership, allocation and management, with a potential needed copy. It's really just a vector<uint32_t>. I'm not extremely picky on how it happens, it will just be more changes. On both sides.

> Can we rename this one to VK_VIDEO_NO_STDOUT_INFO?

Sounds good to me.

What about: #if !defined(DISABLE_VK_VIDEO_PARSER_SIMD_OPTIMIZATIONS)?

It would be good to get ExitWrapper.cpp back. That way we get all flexibility. And you want the implementation of the function in cpp, otherwise the pre-processor will copy it at every include.

common/libs/VkCodecUtils/ExitWrapper.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include <cstdio>
+#include <cstdlib>
+
+// Returns true if the caller would need to handle the exit
+// Otherwise, the function will call exit() that would exit the process.
+inline bool safe_exit(int status, const char* function_name, int line_number, const char* description) {
+    printf("Exiting with a critical error from %s:%d - %s (status: %d)\n", function_name, line_number, description, status);
```
srinathkr-nv: I find the "with a critical error" part problematic -- we have cases where the call site is `safe_exit(EXIT_SUCCESS, ...)`.

common/libs/VkCodecUtils/ExitWrapper.h
srinathkr-nv: I kind of agree with @vasn1k here regarding the separate .cpp file, although one other header-only approach is for other uses of this code to provide their own ExitWrapper.h with a different definition of `safe_exit()`.

common/libs/VkCodecUtils/ExitWrapper.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include <cstdio>
+#include <cstdlib>
+
+// Returns true if the caller would need to handle the exit
+// Otherwise, the function will call exit() that would exit the process.
+inline bool safe_exit(int status, const char* function_name, int line_number, const char* description) {
```
srinathkr-nv: Why does this return `bool` ? `void` is perfectly good because no caller checks for the return value from `SAFE_EXIT()`.

common/libs/VkCodecUtils/ExitWrapper.h
vasn1k: Per our discussion, this morning. Lets get rid of ExitWrapper and convert all exit() calls into returns.
I'll look into compiler settings to disallow exit() entirely later.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderDpbH264.cpp
```diff
@@ -278,7 +279,7 @@ int8_t VkEncDpbH264::DpbPictureEnd(const PicInfoH264 *pPicInfo,
             }
             if (m_currDpbIdx >= MAX_DPB_SLOTS) {
                 VK_DPB_DBG_PRINT(("could not allocate a frame buffer\n"));
-                exit(1);
+                SAFE_EXIT(1, "Failed to allocate frame buffer - DPB slots exhausted");
```
srinathkr-nv: For the elimination of `exit()` from the code, this entire block (lines 280 - 283) can be deleted. We will not have a case where `m_currDpbIdx >= MAX_DPB_SLOTS`  because if it were true, we would have not come out of the `while (IsDpbFull())` loop above. It's probably sufficient to add a comment here stating this while deleting this code.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderDpbH264.cpp
```diff
@@ -355,7 +356,7 @@ int8_t VkEncDpbH264::DpbPictureEnd(const PicInfoH264 *pPicInfo,
                     }
                     if (m_currDpbIdx >= MAX_DPB_SLOTS) {
                         VK_DPB_DBG_PRINT(("could not allocate a frame buffer\n"));
-                        exit(1);
+                        SAFE_EXIT(1, "Failed to allocate frame buffer - DPB slots exhausted");
```
srinathkr-nv: This block of code in lines 357 - 360 can be deleted because this is the `else` branch of an if-else, where the DBP is not full, so the above for loop (lines 354 - 356) will find an empty entry. Just like I wrote in the previous comment, it's probably sufficient to leave a comment here stating this while deleting this code (which will eliminate the `exit()`).

srinathkr-nv:
The latest changes (as of commit e514b85) LGTM.

zlatinski:
I'm merging these basic changes and compiler fixes so far. There are going to be follow-up changes from https://github.com/nvpro-samples/vk_video_samples/pull/140

PR #147, opened 2025-06-13
Vulkan Video VP9 decode support

zlatinski:
Add support for Vulkan Video VP9 decode

The base changes required to support VP9 decode have been added.
Added VP9-decode-specific changes

Unfortunately, the Vulkan SDK with VP9 is not available yet.
The latest version of the Vulkan (video) headers is required.
Please make sure to update the headers by removing the old ones.
rm -rf ./build/_deps/vulkan-headers*


krajunv (approved):
LGTM

PR #148, opened 2025-06-16
Remove exits that will end the process

vasn1k:


dabrain34:
Can you describe a use case where this patch is necessary ? Is it because of the decoder library ?

dabrain34:
Thank you for the confirmation. I was asking because it would be nice to handle properly the result of decode config in the application by for example an exit(success) if `--help` or `--version` are invoked.

PR #155, opened 2025-08-06
Add intra refresh support

srinathkr-nv:
This set of changes adds support in the encoder sample application for the [VK_KHR_video_encode_intra_refresh](https://registry.khronos.org/vulkan/specs/latest/man/html/VK_KHR_video_encode_intra_refresh.html) extension.

### DIFF
--- a/common/include/VkVideoCore/VulkanVideoCapabilities.h
+++ b/common/include/VkVideoCore/VulkanVideoCapabilities.h
@@ -68,9 +68,11 @@ public:
                                                VkVideoEncodeCapabilitiesKHR& videoEncodeCapabilities,
                                                VkVideoEncodeCodecCapabilitiesKHR& videoCodecCapabilities,
                                                VkVideoEncodeQuantizationMapCapabilitiesKHR& quantizationMapCapabilities,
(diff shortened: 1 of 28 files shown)


srinathkr-nv:
These commits have been cherry-picked from https://github.com/KhronosGroup/Vulkan-Video-Samples and are identical to the ones posted for review in https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/79 . I have intentionally created separate commits so that certain changes can be reverted cleanly, if we find out later that they regressed something. Also, preserving the commit structure will make future porting of changes from/to https://github.com/KhronosGroup/Vulkan-Video-Samples easier.

The first commit (Add missing 'override' ...) is a bug fix which I noticed when working on these changes, so I can split it off into a separate PR. The changes for supporting multiple slices are a prerequisite for (some modes of) intra-refresh to work, but I can split it off into a separate PR too. Two of the last three commits introduce a couple of "test-only" options to the encoder sample application, so I can create a separate PR for that too.

In summary, to make review easier, I can separate out the changes into 4 PRs:
- the `override` qualifier bugfix
- support for encoding multiple slices
- actual intra-refresh support
- addition of test-only code

PR #158, opened 2025-09-11
encode: Set the MSB shift for 10/12-bit input

zlatinski:
Set automatically the input.msbShift value to the default shift, if not explicitly set by the API for 10/12-bit input.

### DIFF
--- a/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h
@@ -1009,6 +1009,27 @@ public:
             return VK_ERROR_INVALID_VIDEO_STD_PARAMETERS_KHR;
         }
 
+        // Deal with the input shift values, if not explicitly set.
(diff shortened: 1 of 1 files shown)


vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h
```diff
@@ -1009,6 +1009,25 @@ struct EncoderConfig : public VkVideoRefCountBase {
             return VK_ERROR_INVALID_VIDEO_STD_PARAMETERS_KHR;
         }
 
+        if (input.msbShift == -1) {
+
```
srinathkr-nv, line 1014: Blank lines like the one here and on lines 1015, 1023, 1025 and 1027 seem unnecessary. Otherwise this change LGTM.

zlatinski: Oho, these lines are left like this intentionally for readability reasons. If you pull this out in your source tree, you'd see what I'm talking about

PR #159, opened 2025-09-16
Update DPB after populating reference slot infos

krajunv:
Populating reference slot details for encode after DPB management and picture marking sometime result into incorrect details.   This includes incorrect picture, poc and other values related to the reference picture.  So, populate all the details before DPB management.


zlatinski (commented):
Thank you, Raju! I think we need a more elaborate description of this change on why and how.

krajunv:
> Thank you, Raju! I think we need a more elaborate description of this change on why and how.

Sure @zlatinski, I will update the description with detailed information.

PR #221, opened 2026-04-02
common: do not modify output filename

dabrain34:
just inform user that the output filename is not correct but do not modify it.

If the name is changed, fluster can not verify that the ouput file is correct as it is expecting ".out" extension.

### DIFF
--- a/common/libs/VkCodecUtils/VkVideoFrameToFile.cpp
+++ b/common/libs/VkCodecUtils/VkVideoFrameToFile.cpp
@@ -191,21 +191,16 @@ public:
             m_outputFile = nullptr;
         }
 
-        std::string fileNameWithModExt;
         // Check if the file does not have a y4m extension,
         // but y4m format is requested.
         if (y4mFormat && !hasExtension(fileName, ".y4m")) {
             std::cout << std::endl << "y4m output format is requested, ";
             std::cout << "but the output file's (" << fileName << ") extension isn't .y4m!"
                       << std::endl;
-            fileNameWithModExt = fileName + std::string(".y4m");
-            fileName = fileNameWithModExt.c_str();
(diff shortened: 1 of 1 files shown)


common/libs/VkCodecUtils/VkVideoFrameToFile.cpp
```diff
@@ -191,21 +191,16 @@ class VkVideoFrameToFileImpl : public VkVideoFrameOutput {
...
                       << std::endl;
-            fileNameWithModExt = fileName + std::string(".y4m");
-            fileName = fileNameWithModExt.c_str();
         } else if ((y4mFormat == false) && !hasExtension(fileName, ".yuv")) {
             std::cout << std::endl << "Raw yuv output format is requested, ";
             std::cout << "but the output file's (" << fileName << ") extension isn't .yuv!"
                       << std::endl;
-            fileNameWithModExt = fileName + std::string(".yuv");
-            fileName = fileNameWithModExt.c_str();
         }
```
zlatinski, line 204: I would rather add a check for .out and not modify it in that case.

zlatinski (commented):
Can we please just extend the checks?

zlatinski:
No respond for  awhile. Closing this issue. Reopen if needed, please.

common/libs/VkCodecUtils/VkVideoFrameToFile.cpp
dabrain34: Following your recommendation, I changed to not modify  only with .out extenstion

dabrain34:
unfortunately I can not reopen this PR. I pushed to my previous branch, a patch to follow your recommendation.

I think it could be nice to be able to run fluster with `vk_video_samples`

PR #225, opened 2026-04-30
Rename H.265 aligned picture size parameters for accuracy

krajunv:
Rename the output parameters of GetCtbAlignedPicSizeInSamples so their names reflect the actual values they carry: aligned picture width and height in samples, not CTB counts.

No functional change.


krajunv:
@zlatinski @srinathkr-nv Please check the renaming of the variables so that they reflect the actual values they carry.

srinathkr-nv (approved):
Yes, I think this change makes sense.

PR #226, opened 2026-05-04
Fix y4m 422 output

vasn1k:
The Y4M formatter did not have the necessary output flag for the 4:2:2 format. This would prevent viewer software from opening the Y4M correctly.


zlatinski (approved):
LGTM - thank you!

PR #228, opened 2026-05-08
Bitstream assembly refactor

krajunv:
Split VkVideoEncoder::AssembleBitstreamData into ReadbackBitstreamData
and WriteBitstreamToFile so the AV1 path can reuse the same code as
the base encoder and the async-assembly worker thread, and rewrite
VkVideoEncoderAV1::AssembleBitstreamData on top of two new helpers:
  - BuildFrameObuSequence: stages sequence-header + frame OBUs into
    m_bitstream[frameIdx].
  - FlushBatchedTemporalUnit: emits IVF headers, the TD OBU, and all
    batched per-frame payloads, then releases the per-slot capacity.

Drive-by cleanups:
  - Add encodeInfo.dstBufferOffset to the GPU pointer when reading
    back the bitstream so the offset is honored even if it ever
    becomes nonzero.
  - Hoist the AV1 temporal-delimiter OBU into a class-scope constant
    (TD_OBU_HDR / TD_OBU_PAYLOAD_SIZE) instead of redeclaring the
    two-byte literal at each call site.
  - Drop the noisy "[EncoderConfig] Unknown positional args" stderr
    warning in EncoderConfig::ParseArguments.

### DIFF
--- a/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.cpp
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.cpp
@@ -918,14 +918,6 @@ int EncoderConfig::ParseArguments(int argc, const char *argv[])
         return -1;
     }
 
-    if (argcount > 0) {
-        fprintf(stderr, "[EncoderConfig] Unknown positional args (%d) passed to DoParseArguments:", argcount);
-        for (int a = 0; a < argcount; a++) {
-            fprintf(stderr, " '%s'", arglist[a]);
-        }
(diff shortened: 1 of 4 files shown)


dabrain34:
It would be nice to backport this feature to KHR repository.

zlatinski:
The change looks good, but based on the review, I have a few follow-up suggestions for improvements.

Please have a look at: https://github.com/nvpro-samples/vk_video_samples/pull/232

zlatinski:
Alread merged as part of Pr 228+fix+syncfix

PR #229, opened 2026-05-13
Support inline session parameters

vasn1k:
Add support for inline parameters passing to the sample app for decode h264 h265 and av1


common/include/vulkan_interfaces.h
```diff
@@ -19,3 +19,4 @@
 #define VK_ENABLE_BETA_EXTENSIONS 1
 #endif
 #include "vulkan/vulkan.h"
+#include "vulkan/vk_khr_video_maintenance2_compat.h"
```
srinathkr-nv: This header is not part of this change (and isn't even needed anymore).

vk_video_encoder/src/vulkan_video_encoder.cpp
srinathkr-nv, line 1: We don't need to enable VK_KHR_video_maintenance2 when using the encoder.

vk_video_decoder/libs/VkVideoParser/VulkanVideoParser.cpp
```diff
@@ -173,7 +173,7 @@ struct nvVideoAV1PicParameters {
     uint16_t height_in_sbs_minus_1[64];
     StdVideoDecodeAV1PictureInfo stdPictureInfo; // memory for the pointer in pictureInfo
     VkVideoDecodeAV1PictureInfoKHR pictureInfo;
-    VkVideoDecodeAV1SessionParametersCreateInfoKHR pictureParameters;
+    VkVideoDecodeAV1InlineSessionParametersInfoKHR av1InlineSessionParameters {};
```
srinathkr-nv: Nit: it would be better to rename this member to `inlineSessionParameters` for consistency with the rest of the codecs.

common/libs/VkCodecUtils/DecoderConfig.h
```diff
@@ -372,6 +373,14 @@ struct DecoderConfig {
                     noPresent = true;
                     return true;
                 }},
+            {"--inlineParams", nullptr, 0,
+                "Enable VK_KHR_video_maintenance2: inline VPS/SPS/PPS (or AV1 sequence header) on decode "
```
srinathkr-nv: This description is too long. When the help text for the decoder sample app is printed, this description isn't pretty-printed. It appears as one long sentence, which gets split across lines depending on the terminal / console width.

I suggest making this description more concise, like "Use inline session parameters (requires driver support for VK_KHR_video_maintenance2)".

common/libs/VkCodecUtils/VulkanDeviceContext.cpp
```diff
@@ -1299,6 +1333,18 @@ VkResult VulkanDeviceContext::InitVulkanDecoderDevice(const char * pAppName,
         VK_KHR_TIMELINE_SEMAPHORE_EXTENSION_NAME,
         nullptr
     };
+    static const char* const requiredDeviceExtensionInlineParams[] = {
```
srinathkr-nv: We don't need this separate definition. The only purpose it serves is to add `VK_KHR_video_maintenance2` in the list of required device extensions, and we can do that in a different way.

vk_video_decoder/include/vkvideo_parser/VulkanVideoParserIf.h
```diff
@@ -508,6 +508,8 @@ typedef struct VkParserInitDecodeParameters {
 
     // If set, Picture Parameters are going to be provided via UpdatePictureParameters callback
     bool outOfBandPictureParameters;
+    // If true, use VK_KHR_video_maintenance2 inline session parameters on decode (requires device support).
+    bool inlineSessionParameters;
```
srinathkr-nv, line 510: This can replace `outOfBandPictureParameters` instead of being another entry, because `inlineSessionParameters == !outOfBandPictureParameters`.

vk_video_decoder/include/vkvideo_parser/VulkanVideoParserParams.h
```diff
@@ -56,6 +53,10 @@ struct VkParserPerFrameDecodeParameters {
     int32_t numGopReferenceSlots;
     int8_t pGopReferenceImagesIndexes[MAX_DPB_REF_AND_SETUP_SLOTS];
     VkVideoPictureResourceInfoKHR pictureResources[MAX_DPB_REF_AND_SETUP_SLOTS];
+
+    // Optional pNext chain for VkVideoBeginCodingInfoKHR (legacy in-band session parameters).
+    const void* beginCodingInfoPictureParametersExt;
```
srinathkr-nv: It doesn't look like `beginCodingInfoPictureParametersExt` is used anymore in the parser code, so it can be removed.

vk_video_decoder/libs/VkVideoDecoder/VkVideoDecoder.cpp
```diff
@@ -275,6 +275,9 @@ int32_t VkVideoDecoder::StartVideoSequence(VkParserDetectedVideoFormat* pVideoFo
         sessionCreateFlags |= VK_VIDEO_SESSION_CREATE_INLINE_QUERIES_BIT_KHR;
     }
 #endif // VK_KHR_video_maintenance1
+    if (m_vkDevCtx->GetInlineSessionParametersEnabled()) {
```
srinathkr-nv: We can instead use the `GetVideoMaintenance2FeatureSupported()` method added in this PR, because that method isn't being used anywhere else.

vk_video_decoder/libs/VkVideoParser/VulkanVideoParser.cpp
```diff
@@ -668,12 +668,11 @@ class VulkanVideoParser : public VkParserVideoDecodeClient,
         uint32_t defaultMinBufferSize,
         uint32_t bufferOffsetAlignment,
         uint32_t bufferSizeAlignment,
-        bool outOfBandPictureParameters,
         uint32_t errorThreshold);
 
     VulkanVideoParser(VkVideoCodecOperationFlagBitsKHR codecType,
         uint32_t maxNumDecodeSurfaces, uint32_t maxNumDpbSurfaces,
-        uint64_t clockRate);
+        uint64_t clockRate, bool inlineSessionParameters);
```
srinathkr-nv: Why is this being specified in the `VulkanVideoParser` constructor rather than being passed to Initialize() (and replacing `outOfBandPictureParameters`) ?

`outOfBandPictureParameters == true` means "use VkVideoSessionParametersKHR object", and we need the inverse behaviour.

vk_video_decoder/libs/VkVideoParser/VulkanVideoParser.cpp
```diff
@@ -2528,6 +2517,21 @@ bool VulkanVideoParser::DecodePicture(
         pDecodePictureInfo->viewId = 0; // @review: Doesn't seem to be used in Vulkan?
 
         pPictureInfo->sType = VK_STRUCTURE_TYPE_VIDEO_DECODE_AV1_PICTURE_INFO_KHR;
+
+        if (pin->pStdSps) {
```
srinathkr-nv: Is there any case where this would be false? A valid `pStdSps` (i.e. a sequence header) is always needed for decoding.

common/libs/VkCodecUtils/VulkanDeviceContext.cpp
```diff
@@ -1268,8 +1300,10 @@ VkResult VulkanDeviceContext::InitVulkanDecoderDevice(const char * pAppName,
                                                       bool enableValidation,
                                                       bool enableVerboseValidation,
                                                       bool enbaleVerboseDump,
+                                                      bool enableInlineSessionParameters,
                                                       const char * pCustomLoader)
 {
+    m_inlineSessionParametersEnabled = enableInlineSessionParameters ? 1u : 0u;
```
srinathkr-nv: Passing in `enableInlineSessionParameters` into this code for use below seems fine to me. However, the Vulkan device context does not seem like the correct place to store information about whether inline session parameters are being used.

common/libs/VkCodecUtils/VulkanDeviceContext.cpp
```diff
@@ -1326,7 +1372,11 @@ VkResult VulkanDeviceContext::InitVulkanDecoderDevice(const char * pAppName,
     }
 
     // Add the Vulkan video required device extensions
-    AddReqDeviceExtensions(requiredDeviceExtension);
+    if (enableInlineSessionParameters) {
```
srinathkr-nv, line 1333: The original `AddReqDeviceExtensions()` call can stay. The only additional piece we need is:

```
if (enableInlineSessionParameters) {
    AddReqDeviceExtension("VK_KHR_video_maintenance2");
}
```

vk_video_decoder/libs/VkVideoDecoder/VkVideoDecoder.cpp
```diff
@@ -1242,65 +1245,71 @@ int VkVideoDecoder::DecodePictureWithParameters(VkParserPerFrameDecodeParameters
     frameSynchronizationInfo.syncOnFrameCompleteFence = true;
     frameSynchronizationInfo.imageSpecsIndex = m_imageSpecsIndex;
 
-    VkSharedBaseObj<VkVideoRefCountBase> currentVkPictureParameters;
```
srinathkr-nv, line 1245: I don't understand the purpose of this diff. Is it a real functional change necessary to support use of inline session parameters, or is it a cleanup? If it is a cleanup, please pull it out into a separate commit and explain its purpose (i.e. what the cleanup achieves).

vk_video_decoder/libs/VkVideoParser/VulkanVideoParser.cpp
```diff
@@ -967,7 +963,10 @@ VkResult VulkanVideoParser::Initialize(
 
     nvdp.referenceClockRate = m_clockRate;
     nvdp.errorThreshold = errorThreshold;
-    nvdp.outOfBandPictureParameters = outOfBandPictureParameters;
+    const bool useInlineForCodec = m_inlineSessionParameters &&
+        (m_codecType != VK_VIDEO_CODEC_OPERATION_DECODE_VP9_BIT_KHR);
```
srinathkr-nv: What is the need to check for VP9 here and below? VP9 doesn't have a sequence header, so enabling inline session parameters for VP9 is a no-op, but that doesn't require special checks because it's sufficient for the VP9-specific block below to ignore the variables related to inline session parameters.

srinathkr-nv (changes requested):

vk_video_decoder/libs/VkVideoDecoder/VkVideoDecoder.cpp
vasn1k: I think it just looks worse than it is because of the extra nesting. The change is to prevent all this being setup when inlineSP is enabled. Otherwise it wasn't possible to enable inlineSP for AV1, I then added the H264 H265 checks because having if (inlineSP) { if (!inlineSP) {} } is weird.
Some cleanup shouldn't be an issue in a change.

vk_video_decoder/libs/VkVideoParser/VulkanVideoParser.cpp
vasn1k: whoops I think I removed that then forgot I removed the variable..

vk_video_decoder/libs/VkVideoParser/VulkanVideoParser.cpp
vasn1k: Mostly because VP9 doesn't have inlineSP support. So I thought we specifically had to prevent enabling it for VP9 to keep the code paths consistent.
I'll remove it.

common/libs/VkCodecUtils/VulkanDeviceContext.cpp
vasn1k: I don't really know how to pass the data to VulkanDeviceContext::CreateVulkanDevice in that case. 
I'll just remove the checks in VulkanDeviceContext::CreateVulkanDevice -- seems a little inconsistent to me but if you have a better suggestion I'll add that.

dabrain34:
hello, would be great to have this feature in KHR repository. Could you please backport it when its ready ?

vk_video_decoder/include/vkvideo_parser/VulkanVideoParserParams.h
```diff
@@ -56,6 +53,9 @@ struct VkParserPerFrameDecodeParameters {
     int32_t numGopReferenceSlots;
     int8_t pGopReferenceImagesIndexes[MAX_DPB_REF_AND_SETUP_SLOTS];
     VkVideoPictureResourceInfoKHR pictureResources[MAX_DPB_REF_AND_SETUP_SLOTS];
+
+    // Optional pNext chain for VkVideoBeginCodingInfoKHR (legacy in-band session parameters).
```
srinathkr-nv: This comment refers to the non-standard method of specifying the session parameters inline, and should not be added because the non-standard method is being deleted by this change. Additionally, it's confusing when placed next to the `useInlinedPictureParameters` member (which is related to the proper, standard method of specifying the session parameters inline).

vk_video_decoder/libs/VkVideoDecoder/VkVideoDecoder.cpp
```diff
@@ -1242,65 +1245,71 @@ int VkVideoDecoder::DecodePictureWithParameters(VkParserPerFrameDecodeParameters
...
-
-        decodeBeginInfo.videoSessionParameters = *pOwnerPictureParameters;
 
-        if (m_dumpDecodeData) {
-            std::cout << "Using object " << decodeBeginInfo.videoSessionParameters <<
-                 " with ID: (" << pOwnerPictureParameters->GetId() << ")" <<
-                 " for SPS: " <<  spsId << ", PPS: " << ppsId << std::endl;
+            assert(pCurrFrameDecParams->pStdSps->GetStdType() == StdVideoPictureParametersSet::TYPE_AV1_SPS);
+            // VkParserVideoPictureParameters registers AV1 sequence header at slot 0 (see CreateParametersObject).
+            const uint32_t av1SeqSlotId = 0;
```
srinathkr-nv: Adding this `const` variable here seems redundant. Why not just do the following near line 1309 below?

```
// VkParserVideoPictureParameters registers AV1 sequence header at slot 0 (see CreateParametersObject)
assert(pOwnerPictureParameters->HasAv1PpsId(0));
```

vk_video_decoder/libs/VkVideoDecoder/VkVideoDecoder.cpp
srinathkr-nv: I understand the need for the nesting, but what made it difficult for me to parse this portion of the change is the change in order of the codec-specific cases. If the original order had been retained (AV1, followed by H.264 and HEVC) the diff would have been simpler. However, I've read this portion again and it seems fine to me.

Also, cleanups within a large change are fine, but they should go into their own commit as far as possible (unless the cleanup is tied to some refactoring and cannot be pulled out), with some explanation in the commit message, unless it is genuinely self-explanatory.

srinathkr-nv (approved):

PR #232, opened 2026-07-16
Pr 228+fix+syncfix

zlatinski:
encoder: factor bitstream assembly into shared readback/write helpers - original from Raju Konda
encoder: address review findings for bitstream assembly refactor - proposed fixes of the above change
encoder: release per-frame resources on the synchronous assembly path - issue found during testing


krajunv:
Looks good to me.  Thanks!

PR #234, opened 2026-08-01
VulkanShaderCompiler: compile GLSL through glslang instead of shaderc

zlatinski:
shaderc is a wrapper around glslang adding a stable C API, #include resolution and a SPIRV-Tools optimisation pass - none of which this library uses. Going straight to glslang removes an external dependency without changing what compiles the shaders, and unblocks the encoder's preprocess compute filter in embedders that vendor glslang but not shaderc.

The backend is a seam rather than a swap. VulkanShaderCompiler.h was already backend-agnostic (it holds the compiler as an opaque void*), so it keeps its shape and no caller changes. Behind it:

  VulkanShaderCompilerBackend.h    virtual interface: CompileGlslToSpirv()
  VulkanShaderCompilerGlslang.cpp  glslang backend (default)
  VulkanShaderCompilerShaderc.cpp  shaderc backend (opt-in)
  VulkanShaderCompiler.cpp         facade: refcounted backend + module creation

Create()/Destroy() are defined in each backend translation unit, so exactly one definition exists per build and the facade needs no preprocessor conditionals at all. Only the selected compiler's library is linked.

Backend selection lives in the new cmake/VulkanShaderCompilerBackend.cmake, included by this project and by the external trees that embed VkCodecUtils sources (the AQ library, ThreadedRenderingVk), so all of them resolve it the same way and can migrate independently:

  -DVK_VIDEO_SAMPLES_SHADER_BACKEND=glslang   (default)
  -DVK_VIDEO_SAMPLES_SHADER_BACKEND=shaderc

Two things the port has to get right, both silent if wrong: GetDefaultResources() is declared at GLOBAL scope, not in namespace glslang, and lives in libglslang-default-resource-limits; passing it is mandatory or glslang rejects valid shaders with spurious "too many ..." errors. And EShMsgSpvRules|EShMsgVulkanRules must be set or glslang applies OpenGL semantics and rejects layout(set=) and push constants. There is deliberately no SPIRV library in the link line: glslang 14 merged GlslangToSpv into libglslang and the leftover libSPIRV exports no symbols.

Removed with the shaderc discovery: the from-source FetchContent chain (SPIRV-Headers -> SPIRV-Tools -> glslang -> shaderc) and its WIN32 install rules, a second always-executed block ending in
message(FATAL_ERROR "Could not find or build Shaderc") that made the tree unconfigurable without shaderc, and a stale #include <shaderc/shaderc.hpp> in VulkanVideoUtils.cpp that used no shaderc symbol but forced shaderc headers onto four more targets. Ten link sites are repointed, not eight - vulkan-video-simple-dec and vulkan-video-enc linked shaderc without compiling the compiler at all.

VK_VIDEO_SAMPLES_COMPUTE_FILTER_SUPPORTED is now defined with the other project-wide macros. It gates nothing in this tree today (the guards added by ac529303 were since removed and the filter is unconditionally compiled, enablePreprocessComputeFilter defaulting true), but embedders that do gate on it - the Chromium vendored copy - now agree with the standalone build.

Link the shared glslang, not the static archives
------------------------------------------------
The glslang CONFIG package resolves glslang::glslang to libglslang.a and declares libSPIRV-Tools.a / libSPIRV-Tools-opt.a as transitive interface dependencies, because that glslang is built with ENABLE_OPT=ON. We compile with SpvOptions::disableOptimizer = true and never call into SPIRV-Tools, so all of it was dead weight: every binary that compiles VulkanShaderCompiler grew by roughly 8 MB - vk_filter_test 506 KB -> 9063 KB, vk-video-dec-test 824 KB -> 9384 KB.

shaderc was linked shared (libshaderc_shared.so.1), and shared glslang is what browser embedders link, so preferring the shared library keeps the deployment shape the port was meant to preserve. find_library now runs first and resolves libglslang.so ahead of the archive; the CONFIG package remains the fallback for SDK layouts that ship no bare libraries. -DVK_VIDEO_SAMPLES_GLSLANG_STATIC=ON selects the archives deliberately.

The shared libraries are complete for our use: libglslang.so.15.3.0 exports both GlslangToSpv overloads, libglslang-default-resource-limits.so.15.3.0 exports GetDefaultResources, and libglslang.so has no SPIRV-Tools DT_NEEDED, so nothing extra is pulled in at runtime either.

Windows / Vulkan SDK discovery
------------------------------
Searching only the default library paths is enough on Linux, where distributions install glslang system-wide, but the Vulkan SDK is not on any default path - so configuring on Windows failed outright with "glslang not found". Three Windows-specific things were missing:

  - Vulkan SDK hints. Both the glslang and the shaderc branch now look in $VULKAN_SDK/Lib and /Include (and the lowercase spellings used by the Linux SDK). The shaderc branch had the same gap, so the opt-in backend was equally unbuildable there.
  - Debug/release halves. MSVC cannot mix a release-CRT library into a debug build and the SDK ships the debug ones under a "d" suffix, so glslangd and glslang-default-resource-limitsd are located and applied with the debug/optimized keywords.
  - SPIRV-Tools. The SDK's glslang is a static library built with ENABLE_OPT=ON, so it carries unresolved spvContextCreate / spvValidatorOptions* references that the linker must satisfy even though we never enable the optimizer. These are linked on MSVC only; the Linux shared library resolves them internally and needs none of it. This is also why the "do not link SPIRV" note above stays correct: SPIRV.lib is a 2 KB stub in the SDK exactly as libSPIRV.so is on Linux, and it is SPIRV-Tools - a different project - that is actually required.

Also found while testing, not fixable from inside the project: the SDK's glslang references C++23 STL helpers (__std_find_last_trivial_1, __std_search_1, __std_remove_8) that exist only in MSVC toolset 14.40+, while Visual Studio's default for v143 on this rig is 14.36. CMake locks the toolset in before reading any CMakeLists, so configure with VCToolsVersion=14.44.35207 in the environment - the same workaround ThreadedRenderingVk's SelectMsvcToolset.cmake documents for shaderc_combined. Note that -T version=... is rejected by CMake 3.26 here; the environment variable is what works.

Verification
------------
Linux, NVIDIA RTX 5080, Release, 0 warnings: sizes are back to baseline (vk_filter_test 512 KB, vk-video-dec-test 834 KB, both within 10 KB of the shaderc build) and ldd shows libglslang.so.15 plus libglslang-default-resource-limits.so.15 with zero shaderc symbols. All five A/B signals match the shaderc baseline byte for byte: vk_filter_test 49/49, h264/hevc/av1 3-plane I420 and h264 2-plane encodes, and the decoder --enablePostProcessFilter output. Re-confirmed after the Windows changes on driver 620.05, and with the shaderc backend selected, which reproduces the same five signals and links only libshaderc_shared.

Windows VM (RTX 5080, VS 2022 + toolset 14.44, SDK 1.4.341.1, Release): both backends build with zero unresolved externals, vk_filter_test is 49/49 on each, and the h264/hevc/av1 3-plane I420 encodes are byte identical between the glslang and shaderc backends - and identical to the Linux bitstreams for the same GPU. Linux discovery is unchanged: still the shared libglslang.so, with none of the MSVC-only paths taken.


zlatinski:
This has already been submitted as part of https://github.com/nvpro-samples/vk_video_samples/pull/235.

PR #237, opened 2026-08-18
New encoder formats and filter tests fixes

zlatinski:


vk_video_encoder/src/vulkan_video_encoder_ext.cpp
```diff
@@ -163,21 +166,49 @@ VkResult VulkanVideoEncoderExtImpl::BuildEncoderConfig(
...
-        argStrings.push_back("--inputBpp");
-        argStrings.push_back(std::to_string(bpp));
+    // Input geometry: derive bit depth, chroma subsampling AND plane count from
+    // the VkFormat.
+    //
+    // This is the only consumer of VkVideoEncoderConfig::inputFormat. It used to
+    // switch on the format purely to set --inputBpp for P010/P012, leaving
+    // chromaSubsampling and numPlanes at their 4:2:0 defaults -- which pinned every
+    // caller of this library to 4:2:0 regardless of the format requested. Measured
+    // 2026-08-13: a 4:4:4 or 4:2:2 request encoded as yuv420p and exited 0.
```
srinathkr-nv: The last sentence "Measured 2026-08-13: ..." should be removed because it is no longer relevant to the current state of the code.

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
```diff
@@ -593,9 +593,37 @@ VkResult VkVideoEncoder::SetExternalInputFrame(
...
+            // -> CopyLinearToOptimalImage, which fetches YcbcrVkFormatInfo() (NULL for
+            // these formats) and dereferences it with no NULL check.
+            case VK_FORMAT_R8G8B8A8_UNORM:                                 // AYUV
+            case VK_FORMAT_A2B10G10R10_UNORM_PACK32:                       // Y410
+                isDirectlyEncodable = (format == m_imageInFormat);
+                break;
+            // P012 removed: there is no 12-bit row in any driver encode-input
+            // table (12-bit 4:2:0 is decode-output only) and no 12-bit
+            // NV_ENC_BUFFER_FORMAT on any chip, so claiming it here only turned an
+            // early clear rejection into a late opaque one.
```
srinathkr-nv: This paragraph can be made more concise by simply saying: "The P012 case has been removed as this format is not supported for encoding. Claiming it as supported here turned an early clear rejection into a late opaque one."

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
```diff
@@ -851,9 +879,19 @@ void VkVideoEncoder::CopyYCbCrPlanesDirectCPU(
     // Get format information
     const VkMpFormatInfo* formatInfo = YcbcrVkFormatInfo(format);
 
+    // Packed 4:4:4 (AYUV / Y410 / Y416) has no multi-planar descriptor, so formatInfo is
+    // NULL for it. Falling through to the 8-bit default below would set bytesPerPixel to
+    // 1 for what is a 4-byte container and copy only width*1 of each width*4 row --
+    // exactly one quarter of every scanline, leaving the rest of the staging image at
+    // whatever it held before. That decoded to a luma average of 30.2 against the 125.4
+    // that the semi-planar and 3-planar 4:4:4 cells produce from the same source.
```
srinathkr-nv: The numbers in the last sentence don't make sense without context. I suggest removing the last sentence and ending the previous sentence as "... leaving the rest of the staging image with stale data."

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
```diff
@@ -1504,6 +1542,15 @@ VkResult VkVideoEncoder::InitEncoder(VkSharedBaseObj<EncoderConfig>& encoderConf
         return result;
     }
 
+    // Re-arm the capacity before the next query. GetVideoFormats takes formatCount by
+    // REFERENCE and overwrites it with min(supportedFormatCount, formatCount)
+    // (VulkanVideoCapabilities.h:377), so the DPB query above shrinks it to the DPB list
+    // length and every later query silently inherits that as its cap. With the DPB list
+    // shorter than the SRC list -- which is the normal case now that SRC advertises both
+    // the semi-planar and the packed 4:4:4 form while the DPB can only ever be
+    // semi-planar -- the trailing SRC entries become invisible and the packed format can
```
srinathkr-nv: The portion "while the DPB can only ever be semi-planar" is not quite accurate because this is a detail of the current implementation, not a universal property of DPB allocations. I suggest rewriting this to "while the DPB is currently only semi-planar"

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h
```diff
@@ -759,6 +775,17 @@ struct EncoderConfig : public VkVideoRefCountBase {
     uint8_t  encodeBitDepthLuma;
     uint8_t  encodeBitDepthChroma;
     uint8_t  encodeNumPlanes;
+    // Prefer the packed 4:4:4 encode-source format (AYUV / Y410) when the driver
+    // advertises both representations for the profile.
+    //
+    // For a 4:4:4 profile the driver lists BOTH the 2-plane form and the packed form,
+    // and the 2-plane one comes first, so anything that takes the driver's first entry
```
srinathkr-nv: The portion "and the 2-plane one comes first" refers to a current implementation detail. I suggest rewriting this to "and the listing order is not guaranteed".

common/libs/tests/inc/FilterTestApp.h
```diff
@@ -131,6 +131,13 @@ struct TestCaseConfig {
 struct TestResult {
     std::string     testName;
     bool            passed{false};
+    // Distinct from passed==false. A case is "unvalidated" when the harness could not
+    // check its pixels at all (e.g. optimal-tiled resources, whose upload and readback
+    // are not implemented), as opposed to having checked and found them wrong. Counting
```
srinathkr-nv: I find the sentence "Counting those as passes ..." a bit unnecessary and prone to quickly going stale once we add one more test case. The important information it is trying to convey is why this new parameter is needed, so this sentence and the last one can be combined concisely as "Such cases are their own category and must be reported separately for the test suite to provide meaningful results."

common/libs/tests/src/ColorConversion.cpp
```diff
@@ -393,64 +393,89 @@ void convertNV12toRGBA(const uint8_t* yPlane,
     }
 }
 
-void convertRGBAtoP010(const uint8_t* rgba,
-                       uint32_t width, uint32_t height,
-                       ColorPrimaries primaries, ColorRange range,
-                       std::vector<uint16_t>& yPlane,
-                       std::vector<uint16_t>& uvPlane)
+// One implementation for every 16-bit-container semi-planar format. P010, P012 and P210
+// differ only in bit depth and chroma subsampling, and three near-copies of this loop is
```
srinathkr-nv: This sentence refers to a loop but it's not clear which loop it is, and seems to refer to an intermediate state of the code which is no longer present here. For clarity, I suggest rewriting this paragraph as:


```suggestion
// One implementation for every 16-bit-container semi-planar format. This avoids
// requiring near-identical and duplicated code for processing P010, P012 and P210
// which differ only in bit depth and chroma subsampling
```

srinathkr-nv (changes requested):

common/libs/tests/src/ColorConversion.cpp
zlatinski: @Thank you, @srinathkr-nv!  All your feedback is addressed now.

PR #241, opened 2026-09-16
Compiler fixes and reorder flag addition

vasn1k:
Fixes a few compiler related warnings and adds the no-reorder flags for future compilation.


common/libs/VkCodecUtils/VulkanFilterYuvCompute.h
```diff
@@ -814,6 +814,10 @@ class VulkanFilterYuvCompute : public VulkanFilter
         , m_workgroupSizeX(16)
         , m_workgroupSizeY(16)
         , m_maxNumFrames(maxNumFrames)
+        , m_inputPackedYcbcr(nullptr)
```
zlatinski, line 817: Not all compilers support this, so we need to always do an explicit member init ordering. Do you have this problem at the ToT?

cmake/LinuxSettings.cmake
```diff
@@ -93,6 +93,11 @@ if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "Clang")
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wimplicit-fallthrough")
     endif()
 
+    check_cxx_compiler_flag(-Werror=reorder COMPILER_SUPPORTS_REORDER_ERROR)
+    if(COMPILER_SUPPORTS_REORDER_ERROR)
```
zlatinski, line 97: Let's please fix all the issues. If all the compilers supported that, it would be okay, but I don't think is the case.

zlatinski (changes requested):
Thank you, Vassili, I prefer we fix the ordering of he members and not use flags which are not supported with all compilers.

zlatinski:
LGTM, thank you for the fixes!

